PR #9, opened 2020-06-16
Box Component

nathanyoung:
A `<Box>` is a layout component to build UIs with consistent padding and spacing between elements.

- It lays out it contents in a specified `direction`
- Renders a `<div />` element by default, but can be changed using `as`
- Allows you to specify the spacing between child elements using `childGap`

It essentially is a component api wrapper for our design tokens.


nathanyoung:
[Box](https://trello.com/c/y1IuRI6O/144-box)

PR #16, opened 2020-06-30
feat: WIP Input component

juanfabrega:
This is **WIP** to review today as we plan out the rest of the form components.


src/components/Input/Input.jsx
```diff
@@ -0,0 +1,141 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import { v4 as uuid } from 'uuid';
+import './Input.scss';
```
JCarmine, line 5: Are we planning on using CSS Modules to limit the scope of these styles? My concern is that we don't know what environment these components will end up in. With classnames like `.input` and `.label` it's very possible that these styles can bleed out.

src/components/Input/Input.jsx
```diff
@@ -0,0 +1,141 @@
...
+
+const Input = (props) => {
```
JCarmine, line 7: So, we're constructing an object in the host app, passing this, and using de-structuring to get the values? Why not just pass the explicitly as we're doing in logistics-app?

nathanyoung:
I think we're going to want to allow for an `autocomplete` prop so we can disable it for some of the sign up fields.

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete

nathanyoung:
[Input Component](https://trello.com/c/7z6sNNbp/42-input-component)

src/components/Input/Input.jsx
```diff
@@ -0,0 +1,141 @@
...
+   */
+  id: PropTypes.string,
+  /**
+   * The input's disabled attribute
+   */
+  isDisabled: PropTypes.bool,
+  /**
+   * Determines if input is required or not. (Label will have an asterisk if required)
+   */
+  isRequired: PropTypes.bool,
```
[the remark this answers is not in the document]
juanfabrega: Hey @JCarmine this being a draft PR, you saw it right mid-development. I intend to push an update with more stuff today.

src/components/Input/Input.jsx
```diff
@@ -0,0 +1,141 @@
...
+    label,
+    onBlur,
+    onChange,
+    onFocus,
+    placeholder,
+    type,
+    value,
+  } = props;
+
+  const handleChange = (event) => {
```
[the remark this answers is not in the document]
juanfabrega: I am not seeing linter errors but we might need to dig into the config because those are missing semicolons and I'm pretty sure event gets flagged as a global due to MSIE. I will change these, and then figure out why my linter is not picking it up.

src/components/Input/Input.jsx
```diff
@@ -0,0 +1,141 @@
...
+      placeholder={placeholder}
+      type={type}
+      value={value}
+    />
+  );
+
+  const renderLabel = () => {
+    if (!label) return;
+
+    return (
```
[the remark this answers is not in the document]
juanfabrega: Not sure what's going on here. Again, not seeing this on my end but let's figure it out!

src/components/Input/Input.scss
```diff
@@ -0,0 +1,28 @@
+/* @TODO -- replace with variables or use global utility classes */
+.input {
+  height: 40px;
+  border: 1px solid $color-font-grey-lighter;
+  box-sizing: border-box;
```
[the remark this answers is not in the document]
juanfabrega: Yeah, we need to actually talk about global css in general, since we have little to none, right now.

src/components/Input/Input.scss
```diff
@@ -0,0 +1,28 @@
...
+  &:disabled {
+    color: $color-font-grey-light;
+  }
+}
+
+.label {
+  font-size: $size-font-sm;
+  color: $color-font-base;
+  display: block;
+  font-weight: 700;
```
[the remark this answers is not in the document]
juanfabrega: Yes, but we don't have it yet, we will be adding more tokens this week.

src/components/Input/Input.jsx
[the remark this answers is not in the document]
nathanyoung: I'm seeing the errors too.
Screen Shot 2020-07-01 at 7 46 22 AM

src/components/Input/Input.jsx
```diff
@@ -0,0 +1,141 @@
...
+   */
+  onFocus: PropTypes.func,
+  /**
+   * The input placeholder attribute
+   */
+  placeholder: PropTypes.string,
+  /**
+   * The input 'type' value. Defaults to type 'text'.
+   */
+  type: PropTypes.oneOf(['text', 'password', 'email', 'tel', 'url', 'search']),
```
nathanyoung, line 122: Thinking about the name of the component, `Input` doesn't seems to be specific enough considering these are all text based types.

src/components/Input/Input.jsx
```diff
@@ -0,0 +1,141 @@
...
+    <input
+      className={classes}
+      disabled={isDisabled}
+      id={inputId}
+      onBlur={handleBlur}
+      onChange={handleChange}
+      onFocus={handleFocus}
+      placeholder={placeholder}
+      type={type}
+      value={value}
```
nathanyoung, line 54: should add `required aria-required="true"` attributes if `isRequired`

https://dequeuniversity.com/tips/identify-required-elements

ericyliu:
1) @nathanyoung can i get access to the trello board? I can't view the Input Component ticket
2) have we considered [styled-components](https://styled-components.com/) for our css? that'll give us scoping for css classes so we dont have generic classes like `.input`

src/components/Input/Input.stories.jsx
```diff
@@ -0,0 +1,59 @@
...
+storiesOf('Input', module)
+  .add('Basic Text Input', () => {
+    const basicInputValue = 'Hello World!';
+    const disabledInputValue = 'I am disabled';
+    const placeholderInputValue = '';
+    const withLabelInputValue = 'I have a label!';
+
+    return (
+      <>
+        <div style={{ marginBottom: '1rem' }}>
```
ericyliu: should we make this div of `{ marginBottom: '1rem' }` a reusable component?? something like a `StoryComponentContainer`

src/components/Input/Input.jsx
```diff
@@ -0,0 +1,141 @@
...
+    return (
+      <label className="label" htmlFor={inputId}>
+        {label}<span className="font-color-danger font-size-sm">&nbsp;*</span>
+      </label>
+    );
+  }
+
+  return (
+    label ? (
+      <>
```
ericyliu: thats a cool trick, didnt know about this (is it just a div)?

src/components/Input/Input.jsx
```diff
@@ -0,0 +1,141 @@
...
+        {renderLabel()}
+        {renderInput()}
+      </>
+    ) : (
+      <>
+        {renderInput()}
+      </>
+    )
+  );
```
ericyliu, line 95: why not 
```
return {
  <>
    {label ? renderLabel() : null}
    {renderInput()}
  </>
};
```

src/components/Input/Input.jsx
```diff
@@ -0,0 +1,141 @@
...
+  isDisabled: false,
+  isRequired: false,
+  label: undefined,
+  onBlur: undefined,
+  onFocus: undefined,
+  placeholder: '',
+  type: 'text',
+};
+
+export default Input;
```
ericyliu, line 184: i see you're declaring the 2 sub-render functions inside the functional component itself. Is there an advantage to this versus declaring those functions outside of the main component and passing in the relevant info? In functional components, the function gets run every time the component rerenders. We probably don't want to redeclare these functions right?

src/components/Input/Input.jsx
```diff
@@ -0,0 +1,184 @@
...
+  onChange,
+  onFocus,
+  placeholder,
+  type,
+  value,
+}) => {
+
+  const handleChange = (e) => {
+    onChange(e);
+  };
```
ericyliu, line 29: does this need to be wrapped in another function?

src/components/Input/Input.jsx
juanfabrega: @ericyliu this one is a fragment actually, it will just disappear on render so only your elements go in the dom.

src/components/Input/Input.jsx
juanfabrega: Updated this morning, you can review the latest now.

src/components/Input/Input.jsx
juanfabrega: @ericyliu definitely doesn't need to be wrapped at the moment, however I've found these handlers if any refactoring is required in the future, E.G, adding any additional logic on an input change, that would happen in parallel to the callback. E.G: Applying a class, etc...

You are correct that it is useless at the moment, I just wrote it that way for some consistency across the component.

src/components/Input/Input.stories.jsx
juanfabrega: For sure! I kinda slapped this together to get a story going but as we hadd more stories it will be beneficial to figure out the best way to display these. We also will have access to utility classes from our token library so in the future we might just do something like `<div className="m-b-sm">` (margin-bottom-small) but we're in the process of building all these libraries right now.

src/components/Input/Input.jsx
JCarmine: @ericyliu Not sure why we would need to default to `null`. If label is truthy, function fires and label renders. If not, nothing happens.

ericyliu: ah youre right that's better, I didn't see your comment right above this, silly me

src/components/Input/Input.jsx
ericyliu: thanks for the explanation! I always prefer to keep things as simple as possible, but I appreciate consistency :+1:

src/components/Input/Input.jsx
```diff
@@ -0,0 +1,178 @@
...
+      name,
+      onBlur: handleBlur,
+      onChange: handleChange,
+      onFocus: handleFocus,
+      placeholder,
+      type,
+      value,
+    };
+
+    return !getInputMask() ? <input {...inputProps} /> : <Cleave {...inputProps} options={getInputMask()} />;
```
JCarmine, line 77: At a higher level, I'm wondering if `TextInput` should be the name of this component. Also wondering if PhoneInput or CreditCardInput should be their own components? If we can inherit from TextInput efficiently, I believe this would reduce cognitive complexity for the consumer.

At a lower level, should we be checking the `inputMask` prop on this line? And once again, we don't need to check a value in `getInputMask()` because it won't be able to fire this function unless `inputMask` is truthy.

src/components/Input/Input.jsx
```diff
@@ -0,0 +1,178 @@
...
+    {
+      // [`font-size-${size}`]: size,
+      // [`font-color-${color}`]: color,
+    }
+  );
+
+  const inputId = id || uuid();
+
+  const getInputMask = () => {
+    if (!inputMask) return undefined;
```
JCarmine, line 52: This function should never fire if `inputMask` is falsey.

src/components/Input/Input.jsx
```diff
@@ -0,0 +1,178 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import { v4 as uuid } from 'uuid';
+import Cleave from 'cleave.js/react';
+import * as InputMasks from './InputMasks';
+import './Input.scss';
+
+const Input = ({
```
JCarmine, line 9: 👍

src/components/Input/Input.jsx
```diff
@@ -0,0 +1,178 @@
...
+  name,
+  onBlur,
+  onChange,
+  onFocus,
+  placeholder,
+  type,
+  value,
+}) => {
+
+  const handleChange = (e) => {
```
JCarmine, line 27: Think I'm for writing out the word event; it's a little longer, but more obvious. 

The linter error on this line is because only a single argument is being passed to this function, yet it's wrapped in parenthesis. So, this function could look like:

```
const handleChange = event => {
  onChange(e);
};
```

src/components/Input/Input.jsx
```diff
@@ -0,0 +1,178 @@
...
+    if (onBlur) onBlur(e);
+  };
+
+  const inputClasses = classNames(
+    className,
+    'input',
+    {
+      // [`font-size-${size}`]: size,
+      // [`font-color-${color}`]: color,
+    }
```
JCarmine, line 46: Missing trailing comma according to the linter. I'm attempting to get use to the idea of having to add this, but can go either way (leaning towards not having a hanging comma) on whether or not we want to enforce this.

src/components/Input/Input.jsx
JCarmine: Interesting insight; hadn't realized there was a performance impact for doing this. 

I've always been a fan of nesting inside the parent component since they belong to only this component. So for me, it's more about clarity for the dev, as apposed to worrying about a small performance hit. Lastly, the react js docs seem to be for nesting as can be seen in this hooks example: https://reactjs.org/docs/hooks-overview.html

Either way we go on this, this is a pattern that we should establish now, and stick to it.

src/components/Input/Input.jsx
```diff
@@ -0,0 +1,178 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import { v4 as uuid } from 'uuid';
+import Cleave from 'cleave.js/react';
```
JCarmine, line 5: Neat tool!

ericyliu:
@JCarmine so hooks definitely need to called within the parent function component. It's more the subcomponents (renderLabel and renderInput) in this case that can go outside the main parent component. To me, 1 file represents 1 component, so its ok if a file has 1 main component and several sub components defined at the top level (as long as the main component is always at the top or bottom of the file) (i prefer top)

src/components/Input/Input.jsx
ericyliu: @JCarmine so hooks definitely need to called within the parent function component. It's more the subcomponents (renderLabel and renderInput) in this case that can go outside the main parent component. To me, 1 file represents 1 component, so its ok if a file has 1 main component and several sub components defined at the top level (as long as the main component is always at the top or bottom of the file) (i prefer top)

ericyliu: in addition, hooks that don't need to be in the parent component and only have to do with one of the sub components can go directly into the subcomponent

JCarmine:
On the global/scoped CSS question. Global styles in the traditional sense don't make any sense since the components will be exported with only the styles contained within them. 

I've used CSS Modules in the past and the solution for setting something like box-sizing globally would involve using sass, composes, a css var, or just remembering to set it manually in every component.

The solution could be to somehow build a template generator, that when invoked with a component name, generates a folder that contains boilerplate Component, storybook, sass files, etc. This way, we could ensure that the box-sizing rule is included in the boilerplate by default.

src/components/Input/Input.stories.jsx
ericyliu: :+1:

src/components/Input/Input.jsx
juanfabrega: I am with @ericyliu on this one. Typically I like to keep (sub)render functions inside the parent component scope, purely due to the ease of dev experience given that we can keep a single prop scope across the entire thing. In general I would opt for that pattern moving forward but this case is different because the specific render functions actually can be abstracted to their own components. The truth is we will need to re-use the form label for every single input type so it is worth doing that now and having it be agnostic of it's input altogether. 

I will work on updating this tonight and we can review in the morning.

src/components/Input/Input.jsx
juanfabrega: I will definitely rename this to TextInput. I like it much better. In terms of having a PhoneInput, and a CreditCard input, I'm still not sold on them reducing cognitive load more than the inputMask, since we're allowing simple string values for the masks. They would be super easy to abstract though so we can just ask the team and go from there. my feeling is that this has the potential to grow to a list of components that share the exact same API but do only very slightly different things, in which case props are doing their job properly.

src/components/Input/Input.jsx
juanfabrega: Yup, I figured we could just use getInputMask consistently everywhere but there is no need to fire it at all so I will just do the conditional check with inputMask

src/components/Input/Input.jsx
juanfabrega: Hah I like to too actually. I'm generally a fan of long variable names. It got flagged due to it being an IE global I think. Let's come back to this. Maybe we can just disable this in the linter.

src/components/Input/Input.scss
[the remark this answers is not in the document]
nathanyoung: https://github.com/palmetto/palmetto-components/issues/22

src/components/Input/Input.scss
```diff
@@ -0,0 +1,28 @@
...
+  border: 1px solid $color-font-grey-lighter;
+  box-sizing: border-box;
+  box-shadow: 0px 3px 0px rgba(0, 0, 0, 0.05);
+  border-radius: 8px;
+  padding: 0 $size-spacing-sm;
+  font-size: $size-font-base;
+  color: $color-font-base;
+
+  &::placeholder {
+    opacity: 1; /* Override lower placeholder opacity in Firefox */
```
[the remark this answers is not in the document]
nathanyoung: https://github.com/palmetto/palmetto-components/issues/22

src/components/FormLabel/FormLabel.jsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import './FormLabel.scss';
+
+const FormLabel = ({
```
nathanyoung, line 5: Why not `<Label />`?

src/components/TextInput/TextInput.jsx
```diff
@@ -0,0 +1,174 @@
...
+import FormLabel from '../FormLabel/FormLabel';
+
+const getInputMask = (inputMask, availableInputMasks) => {
+  if (typeof inputMask === 'string') {
+    return availableInputMasks[inputMask];
+  }
+
+  return inputMask;
+};
+
```
nathanyoung, line 17:
```suggestion
/**
 * Use TextInput to show where users can enter text based data. It does not maintain any internal state, so its value should be managed by the parent.
 */
```

Add description for documentation. Feel free to alter if this is right.

.eslintrc
```diff
@@ -19,6 +18,8 @@
     "import/no-extraneous-dependencies": [
       "error",
       { "devDependencies": [".storybook/**", "stories/**", "**/*.stories.jsx"] }
-    ]
+    ],
+    "react/jsx-props-no-spreading": [0],
+    "max-len": [1, 120]
   }
 }
```
ericyliu, line 25: are we making it a requirement that all files have a newline at the bottom?

src/components/FormLabel/FormLabel.jsx
juanfabrega: Not sure, figured we should make it specific, but I don't feel super strongly either way. Figured label might be too generic, but maybe it's not.

.eslintrc
juanfabrega: Yes! my editor's linter was crashed yesterday and as a result we missed a bunch of these. I will fix them all, thanks!

src/components/TextInput/TextInput.jsx
```diff
@@ -0,0 +1,174 @@
...
+      )}
+    </>
+  );
+};
+
+Input.propTypes = {
+  /**
+   * The input's 'autocomplete' attribute
+   */
+  autoComplete: PropTypes.oneOf(['on', 'off']),
```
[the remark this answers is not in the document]
juanfabrega: @nathanyoung despite the verbosity, I feel enum is a good way to go here to make sure the api is transparent between our lib and the underlying html.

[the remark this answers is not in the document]
juanfabrega: Actually, I regret everything. Let's do your approach, there are waaay to many options for this.

Issue #17, opened 2020-06-30
Checkbox Component


juanfabrega:
We need a controlled checkbox component. It should expose all standard checkbox properties and events to the consumer, while maintaining a consistent aesthetic, matching this:

<img width="284" alt="Screen Shot 2020-06-30 at 2 46 01 PM" src="https://user-images.githubusercontent.com/19932356/86164855-7402c280-bae0-11ea-85fb-5d5a7e6bdb99.png">

nathanyoung:
![](https://github.trello.services/images/mini-trello-icon.png) [Checkbox Input](https://trello.com/c/k8m4LNnB/48-checkbox-input)

nathanyoung:
https://github.com/palmetto/palmetto-components/pull/66

PR #23, opened 2020-07-02
48 Checkbox Input Component

JCarmine:


src/components/CheckboxInput/CheckboxInput.scss
```diff
@@ -0,0 +1,18 @@
+.checkboxInput {
+  display: flex;
+  align-items: center;
+
+  &.isDisabled {
```
nathanyoung, line 5: Would it be better to use the `:disabled` pseudo class instead of creating a named nested css class?
```suggestion
  &:disabled {
```


https://developer.mozilla.org/en-US/docs/Web/CSS/:disabled

src/components/CheckboxInput/CheckboxInput.jsx
```diff
@@ -0,0 +1,75 @@
...
+   */
+  id: PropTypes.string,
+  /**
+   * Additional classes to add
+   */
+  className: PropTypes.string,
+  /**
+   * Set the checkbox as selected
+   */
+  selected: PropTypes.bool,
```
nathanyoung, line 60: For other boolean props we have `isDisabled` and `isRequired`. Should this be `isSelected`?

src/components/CheckboxInput/CheckboxInput.jsx
```diff
@@ -0,0 +1,75 @@
...
+        id={inputId}
+        checked={isSelected}
+        disabled={isDisabled}
+        onChange={handleChange}
+        type="checkbox"
+        className={className}
+        value={isSelected}
+        style={{ width: '24px', height: '24px' }}
+      />
+      <span className="checkboxInputInstructions">{children}</span>
```
nathanyoung, line 35: should be a <label> right?

juanfabrega, line 35: Whad do we think about making this a label?

src/components/CheckboxInput/CheckboxInput.scss
JCarmine: Not in this case. Here we're applying this class to a div, since it can't technically be disabled. The class is used to gray out the text, and disable pointer events, so, even if there is a link, it can't be clicked.

The actual input does have the `disabled` attribute applied and it handles disabling itself.

src/components/CheckboxInput/CheckboxInput.jsx
JCarmine: Agreed. And changed.

src/components/CheckboxInput/CheckboxInput.jsx
JCarmine: It should be a label, yes. But in the signup use case, it contains a link to view the terms and conditions. And, we don't want it to toggle the checkbox when the user does so. 

Not sure what to do about this one.

JCarmine: See reply to @nathanyoung 's comment.

src/components/CheckboxInput/CheckboxInput.jsx
```diff
@@ -0,0 +1,75 @@
...
+
+const CheckboxInput = ({
+  id,
+  className,
+  selected,
+  onChange,
+  isDisabled,
+  children,
+}) => {
+  const [isSelected, setIsSelected] = useState(selected || false);
```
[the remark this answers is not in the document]
JCarmine: I feel it is the sole concern of the checkbox to know whether it's selected, or not. Having it rely on a potentially malfunctioning parent component to get its state might be dangerous. Making an update with useEffect to make the relationship between the parent and CheckboxInput component more clear.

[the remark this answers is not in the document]
juanfabrega: Well yes, if the parent fails, the checkbox will fail, but that's the main philosophy behind a controlled component. If we introduce state here creates a disconnect between the parent and the child where the state basically lives in two places potentially.

To think of it a different way, by hiding the failure of the parent to control their state properly, we are introducing potential bugs where a parent believes the checkbox state has been updated, even though it has not. For example, if the checkbox is clicked (onChange fires) and the parent fails to capture new state, the checkbox will still change to checked state, basically giving a false positive to the consumer.

src/components/CheckboxInput/CheckboxInput.jsx
```diff
@@ -0,0 +1,78 @@
...
+  className,
+  isSelected,
+  isDisabled,
+  onChange,
+  children,
+}) => {
+  const [selected, setSelected] = useState(isSelected || false);
+  const inputId = id || uuid();
+
+  useEffect(() => {
```
juanfabrega, line 18: Hey @JCarmine just a heads up, I ran this locally but seems the effect is creating an infinite loop. Let's chat about this tomorrow. I feel we can probably get rid of this altogether, but want to hear what you think.

JCarmine: It's already gone. I'll push soon.

src/components/CheckboxInput/CheckboxInput.jsx
nathanyoung: modern browsers handle this correctly now (they don't toggle the checkbox when a nested link is clicked). 

here's a demo: https://codesandbox.io/s/angry-haslett-m6dl9?file=/index.html

src/components/CheckboxInput/CheckboxInput.jsx
```diff
@@ -0,0 +1,72 @@
...
+  return (
+    <div className={cx('checkboxInput', { isDisabled })}>
+      <input
+        id={inputId}
+        checked={isChecked}
+        disabled={isDisabled}
+        onChange={handleChange}
+        type="checkbox"
+        className={className}
+        style={{ width: '24px', height: '24px' }}
```
juanfabrega, line 30: Since we've got a scss file for this now can we add the styling there and use existing variables as opposed to inline styling? The height and width for this one should be `$size-spacing-2-xl` if I'm not mistaken.

src/components/CheckboxInput/CheckboxInput.jsx
```diff
@@ -0,0 +1,72 @@
...
+      />
+      <label htmlFor={inputId} className="checkboxInputInstructions">{children}</label>
```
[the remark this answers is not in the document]
JCarmine: Ah, yep.

src/components/TextInput/TextInput.scss
```diff
@@ -1,6 +1,6 @@
 @import '../../styles/variables.scss';
 /* @TODO -- replace any remaining harcoded values with variables or use global utility classes */
-.input {
+.TextInput {
```
JCarmine, line 3: CSS Modules might not be necessary. We just need to scope our styles well. Perhaps, for each component, we prefix or suffix the class name with `palmettoComponents` to make library components easier to identify if someone is picking through the DOM in the browser? I don't know, just a half thought.

nathanyoung:
#17

nathanyoung:
[CheckboxInput Component](https://trello.com/c/k8m4LNnB/48-checkboxinput-component)

PR #34, opened 2020-07-02
Install a11y add-on and install for existing components.

JCarmine:


nathanyoung:
[Install a11y add-on to storybook so that it runs on all components](https://trello.com/c/L3EfNsxv/32-install-a11y-add-on-to-storybook-so-that-it-runs-on-all-components)

Issue #14, opened 2020-06-30
Input Component


juanfabrega:
We need a controlled input component. It should expose all standard input properties and events to the consumer, while maintaining a consistent aesthetic, matching this:

![Screen Shot 2020-06-30 at 9 54 46 AM](https://user-images.githubusercontent.com/19932356/86134649-e57a4b00-bab7-11ea-993f-0874af56cb9e.png)

**NOT REQUIRED BUT NICE TO HAVE FOR MVP**
* Input masking (for phone number format for example)

nathanyoung:
![](https://github.trello.services/images/mini-trello-icon.png) [Input Component](https://trello.com/c/7z6sNNbp/42-input-component)

juanfabrega:
PR: https://github.com/palmetto/palmetto-components/pull/16

PENDING: Tests.

PR #38, opened 2020-07-03
Add validation error state to TextInput

nathanyoung:
Screen Shot 2020-07-06 at 8 05 06 AM


src/components/TextInput/TextInput.jsx
```diff
@@ -57,9 +58,10 @@ const TextInput = ({
   const inputProps = {
     'aria-required': isRequired,
     'aria-label': label || name,
+    'aria-invalid': !!error,
     autoComplete: !autoComplete ? 'off' : autoComplete,
     autoFocus,
-    className: inputClasses,
+    className: error ? 'error' : null,
```
juanfabrega, line 64: @nathanyoung the error class should probably be added to the inputClasses object above so we don't lose any custom classes applied by the library user.

something like:
```
const inputClasses = classNames(
    className,
    'TextInput',
    { 'error': error },
  );


const inputProps = {
...
className: inputClasses,
...
};
```

nathanyoung:
[TextInput Component](https://trello.com/c/7z6sNNbp/42-textinput-component)

src/components/InputValidationMessage/InputValidationMessage.jsx
```diff
@@ -0,0 +1,18 @@
...
+import PropTypes from 'prop-types';
+
+/**
+ * Used by form inputs such as TextInput, to display a validation message for an invalid input.
+ */
+
+const InputValidationMessage = ({
+  children,
+}) => (
+  <div className="font-color-danger font-size-sm" style={{ marginTop: 'var(--size-spacing-xs)' }}>{children}</div>
```
nathanyoung, line 11: There is no utility class yet for margin and padding from design tokens, so I'm using the css var so I we don't have to have another css file just for this.

nathanyoung:
#14

src/components/InputValidationMessage/InputValidationMessage.jsx
juanfabrega: @nathanyoung good call, y will create these.

juanfabrega, line 11: we might want to allow for consumers to pass a className prop directly to this component. In cases where a input helper may be using base font color, or just any use case that requires overriding the default.

src/components/FormLabel/FormLabel.jsx
```diff
@@ -1,27 +1,37 @@
...
 import './FormLabel.scss';
 
 const FormLabel = ({
+  error,
   inputId,
   isFieldRequired,
   labelText,
 }) => {
+  const labelClasses = classNames(
+    error ? 'error' : null,
```
JCarmine, line 13: Think this can be written as: `'error': error`

JCarmine, line 13: Think this can be written as: `'error': error`.

JCarmine: Regarding naming conventions: 

Instead of having 'fieldRequired', we do 'isFieldRequired'. So, would we have 'hasError' as opposed to just 'error'?

src/components/FormLabel/FormLabel.scss
```diff
@@ -5,5 +5,9 @@
   color: $color-font-base;
   display: block;
   font-weight: 700;
-  margin-bottom: $size-spacing-sm;
+  margin-bottom: $size-spacing-xs;
+
+  &.error {
+    color: $color-brand-danger-base;
+  }
 }
```
JCarmine, line 13: Need to add new line to end.

src/components/FormLabel/FormLabel.scss
```diff
@@ -5,5 +5,9 @@
   color: $color-font-base;
   display: block;
   font-weight: 700;
```
JCarmine, line 7: In cases where we will eventual have tokens for these rules, should we create css vars in the meantime to make these easier to locate when it's time to replace?

We did this in logistics-app with the colors, so we could easily snap the app to the new style guide.

src/components/TextInput/TextInput.jsx
```diff
@@ -24,6 +25,7 @@ const TextInput = ({
   id,
   inputMask,
   isDisabled,
+  error,
```
JCarmine, line 28: `hasError`?

src/styles/variables.scss
```diff
@@ -1,2 +1,3 @@
 @import '~@palmetto/palmetto-design-tokens/build/scss/variables-color.scss';
 @import '~@palmetto/palmetto-design-tokens/build/scss/variables-size.scss';
+@import '~@palmetto/palmetto-design-tokens/build/css/variables-size.css';
```
JCarmine, line 3: Need to add new line to end.

src/components/FormLabel/FormLabel.scss
juanfabrega: @JCarmine Yes, I like this idea.

src/components/FormLabel/FormLabel.jsx
nathanyoung: @JCarmine @juanfabrega I left off the prefix for `error` because it's not just a boolean. It also accepts a `string` that when defined will make the input appear red.

src/components/TextInput/TextInput.jsx
nathanyoung: @JCarmine @juanfabrega I left off the prefix for error because it's not just a boolean. It also accepts a string that when defined will make the input appear red.

src/components/InputValidationMessage/InputValidationMessage.jsx
nathanyoung: since the current thinking is we don't intend for consumers to directly use this component (just like the FormLabel), I'm going to leave off a className prop for now. If we find we need it later, we can add it then.

PR #44, opened 2020-07-07
49 Button Component

JCarmine:


nathanyoung:
[Button](https://trello.com/c/IRG5Uzsi/49-button)

juanfabrega:
> Min-width needs to be there to handle the case where the button's text is short, like "OK", and the state is loading. In this case, the button needs to be wide enough to render the loading indicator, which is positioned absolutely.
> Screen Shot 2020-07-06 at 10 28 59 PM

@JCarmine @nathanyoung I believe the correct course of action here is to not change the button's width at all with a loading state. This is the case across the board but most apparent specifically with full-width buttons which will look pretty funky if they just shrink to the width of the loading indicator.

src/components/Button/Button.scss
```diff
@@ -0,0 +1,81 @@
...
+
+.Palmetto-Button {
+  position: relative;
+  padding: $size-spacing-sm $size-spacing-md;
+  border: none;
+  border-radius: 8px;
+  color: $color-base-white;
+  background-color: #2F8459; // $color-brand-primary-base;
+  font-weight: 700;
+  letter-spacing: .2px;
```
juanfabrega, line 11: Should we make letter spacing global? Or does the button require something unique?

src/components/Button/Button.stories.jsx
```diff
@@ -0,0 +1,78 @@
...
+import { withA11y } from '@storybook/addon-a11y';
+import Button from './Button';
+
+export default {
+  title: 'Button',
+  component: Button,
+  decorators: [withA11y],
+};
+
+const StatefulButton = ({ children, fullWidth, isDisabled, isLoading }) => {
```
juanfabrega, line 13: Linter will complain about inlining more than 3 items in a destructured object.

src/components/Button/Button.jsx
```diff
@@ -0,0 +1,127 @@
...
+  isLoading,
+  onFocus,
+  onBlur,
+}) => {
+  const inputId = id || uuid();
+
+  const disabled = isLoading || isDisabled;
+
+  const buttonClasses = classNames('Palmetto-Button', className, {
+    'loading': isLoading,
```
juanfabrega, line 24: you don't need the quotes on these unless you are interpolating a variable, or using dashes classnames will read the object key.

src/components/Button/Button.jsx
```diff
@@ -0,0 +1,127 @@
...
+   */
+  isLoading: PropTypes.bool,
+  /**
+   * Callback when button receives focus
+   */
+  fullWidth: PropTypes.bool,
+  /**
+   * Callback when button is pressed
+   */
+  onClick: PropTypes.func.isRequired,
```
juanfabrega, line 112: Does this prop need to be required? I'm thinking of the submit button use case where the user isn't really in charge of their callback (when using formik at least). Might be helpful to keep it as optional for those use cases.

src/components/Button/Button.scss
JCarmine: @nathanyoung added this and it was to deal with letters getting too crowded when they have a bold font weight, I believe.

nathanyoung: don't want to make it global, what @JCarmine said is correct. i added the spacing because the 700 weight font needs some spacing between the letters to help with legibility.

JCarmine: If we haven't implemented our actual base font yet, should we be getting this granular with styling?

nathanyoung: it is here: https://github.com/palmetto/palmetto-components/blob/develop/src/styles/reset.scss#L2

nathanyoung:
Screen Shot 2020-07-07 at 4 48 42 PM

Looking at the 'docs' tab in storybook, and showing the story source doesn't show the stateful button component source, so the code samples aren't useful. Therefore, I think it would be more useful to not rely on a helper component to make stories. 

Based on current functionality, we should have separate stories for:

- Default
- Disabled
- FullWidth
- Loading
- eventHandlers

This also applies to TextInput, I just didn't catch it until now.

juanfabrega:
@JCarmine let's get conflicts resolved when you get a chance. Once we figure out the svg jest issue this is good to go.

PR #45, opened 2020-07-07
feat (CheckboxInput): error and isRequired props

nathanyoung:
Screen Shot 2020-07-06 at 9 51 04 PM

- Adds error state for CheckboxInput
- Uses the `<FormLabel/>` component for the CheckboxInput label instead of custom div. 
- adds `isRequired` prop to CheckboxInput


nathanyoung:
[CheckboxInput Component](https://trello.com/c/k8m4LNnB/48-checkboxinput-component)

nathanyoung:
#17

PR #47, opened 2020-07-07
Text input tests

juanfabrega:
Added basic tests to our TextInput and FormLabel components to test:

1. Prop validation.
2. Core functionality
3. Snapshot variations


juanfabrega:
https://trello.com/c/7z6sNNbp/42-textinput-component

juanfabrega:
[TextInput Component](url)

juanfabrega:
[TextInput Component](https://trello.com/c/7z6sNNbp/42-textinput-component)

src/components/TextInput/TextInput.test.jsx
```diff
@@ -0,0 +1,143 @@
...
+      .toContain('Failed prop type: The prop `onChange`');
+  });
+
+  test('Throws an error if "type" prop is anything other than allowed values', () => {
+    console.error = jest.fn(); // eslint-disable-line no-console
+    render(<TextInput onChange={() => null} value="hello" type="notOnTheList" />);
+    expect(console.error).toHaveBeenCalledTimes(1); // eslint-disable-line no-console
+    expect(console.error.mock.calls[0][0]) // eslint-disable-line no-console
+      .toContain('Failed prop type: Invalid prop `type`');
+  });
```
ericyliu, line 37: we actually dont need these validation texts when we change this to typescript as the code wont even compile if the necessary props aren't passed in

src/components/TextInput/TextInput.test.jsx
```diff
@@ -0,0 +1,143 @@
...
+        id="firstName"
+        label="first name"
+        value="hello"
+        onChange={onChangeMock}
+      />,
+    );
+    const inputElement = screen.getByDisplayValue('hello');
+
+    fireEvent.change(inputElement, { target: { value: 'good bye' } });
+    expect(onChangeMock).toHaveBeenCalledTimes(1);
```
ericyliu, line 57: should you check if it was called with the value?

src/components/TextInput/TextInput.test.jsx
juanfabrega: Hey @ericyliu perhaps I'm misunderstanding, but the goal of these is to guard against somebody flipping a prop requirement. Basically this test breaks to let you know that the prop was required for a reason so best not to remove that requirement without explicit intent. Let me know if that makes sense.

src/components/TextInput/TextInput.test.jsx
juanfabrega: @ericyliu Interestingly this seems to be the recommended approach with RTL. I actually tried checking the value but my synthetic events all come back with target of null. Not sure if this is a quirk of RTL or what but we can definitely keep investigating.

src/components/TextInput/TextInput.test.jsx
ericyliu: ohh good point, yeah that's reasonable

src/components/FormLabel/FormLabel.test.jsx
```diff
@@ -0,0 +1,43 @@
...
+
+  test('Matches the snapshot for component with isRequired prop', () => {
+    const { asFragment } = render(<FormLabel inputId="myId" labelText="hello world" isRequired />);
+    expect(asFragment()).toMatchSnapshot();
+  });
+
+  test('Matches the snapshot for component with hasError prop', () => {
+    const { asFragment } = render(<FormLabel inputId="myId" labelText="hello world" hasError />);
+    expect(asFragment()).toMatchSnapshot();
+  });
```
nathanyoung, line 42: remove

src/components/FormLabel/__snapshots__/FormLabel.test.jsx.snap
```diff
@@ -0,0 +1,34 @@
...
+exports[`FormLabel Matches the snapshot for component with isRequired prop 1`] = `
+<DocumentFragment>
+  <label
+    class="label"
+    for="myId"
+  >
+    hello world
+  </label>
+</DocumentFragment>
+`;
```
nathanyoung, line 34: remove

src/components/Heading/Heading.test.jsx
```diff
@@ -52,28 +52,28 @@ describe('Heading', () => {
...
   });
 
   headingComponentColors.forEach((headingComponent, index) => {
     test(`Matches the snapshot for color ${PALMETTO_BRAND_COLOR_OPTIONS[index]}`, () => {
       const { asFragment } = render(headingComponent);
-      expect(asFragment(headingComponent)).toMatchSnapshot();
+      expect(asFragment()).toMatchSnapshot();
     });
   });
 });
```
nathanyoung, line 79: unit test instead. if you want to do that in a separate pr, that works too.

juanfabrega: @nathanyoung I will do this on a separate PR.

src/components/FormLabel/__snapshots__/FormLabel.test.jsx.snap
juanfabrega: @nathanyoung I have replaced these with unit tests.

PR #59, opened 2020-07-08
[79] add typescript support

### LINKED ISSUE #25: Convert repo to typescript
can use rollup plugins after https://github.com/palmetto/palmetto-components/pull/24 is merged

1. https://github.com/rollup/plugins (babel and typescript)

2. connect typescript types and proptypes: https://github.com/milesj/babel-plugin-typescript-to-proptypes
  - comments can be transpiled too https://github.com/milesj/babel-plugin-typescript-to-proptypes#comments-boolean

ericyliu:
Fixes #25 

1. adds typescript support to building, testing, and storybook
2. unideally, typechecking on build happens after building
3. only some components are converted, will convert the other ones piece-meal in separate prs

*notes*

I tried transpiling from proptypes to typescript types and vice versa. Both have their drawbacks:
1. proptypes to typescript means any non-required prop has the possibility of being null (which we know shouldnt happen if we follow the typescript types)
2. typescript to proptypes (https://github.com/milesj/babel-plugin-typescript-to-proptypes) does not look for typings outside of the component file

The solution I ended up on was declaring both, but typescript makes the proptypes match the typings defined, so we should be kept safe in that regard


ericyliu:
@juanfabrega @nathanyoung let me know what the status on that release is so i can merge this

src/components/CheckboxInput/CheckboxInput.tsx
```diff
@@ -9,7 +9,55 @@ import './CheckboxInput.scss';
...
+   */
+  onChange: PropTypes.func.isRequired,
+  /**
+   * Callback function when input is focused
+   */
+  onFocus: PropTypes.func.isRequired,
+  /**
+   * Custom content to be displayed to right of checkbox. Can be any valid node/tree, anchors, etc.
+   */
+  label: PropTypes.node.isRequired,
```
ericyliu, line 57: @juanfabrega so this is both used for the `aria-label (string | undefined)`, as well as passed into `FormLabel (node)`. we might have to restrict the label typing here to `string` if we want it to satisfy both

juanfabrega: @ericyliu this was updated on `develop` branch a few days ago to take both a string and a node. in the `FormLabel` component. Let me know if you don't see the same on your end.

juanfabrega: @ericyliu oh I see what you are saying now. Might as well follow the same proptype we have on `FormLabel` here as well. Can you chance it?

ericyliu: @juanfabrega thats what im saying, the proptype on formlabel is `node`, but that doesnt work for aria-label

ericyliu:
@juanfabrega should i be concerned about the codecov report

src/components/Button/Button.tsx
```diff
@@ -1,72 +1,60 @@
...
-  tabIndex: undefined,
-  onFocus: undefined,
-  onBlur: undefined,
 };
 
-const Button = ({
+const Button: React.FC<Props> = ({
   id,
-  type,
+  type = 'button',
```
mtblanton, line 57: We're duplicating the default as `'button`' here, we'll want to get rid of either the param initialization or the default prop! I prefer param initialization since it's just a JS feature and you don't need to do any tinkering to get it to work with TS.

ericyliu: youre right! removed the defaultProps object and replaced it with param initialization

src/components/CheckboxInput/CheckboxInput.tsx
```diff
@@ -9,93 +9,72 @@ import './CheckboxInput.scss';
...
   /**
    * Callback function when input is blurred.
    */
-  onBlur: PropTypes.func,
+  onBlur?: (event: React.FocusEvent<HTMLInputElement>) => void;
   /**
    * Callback function when input is changed
    */
-  onChange: PropTypes.func.isRequired,
+  onChange?: (event: ChangeEvent<HTMLInputElement>) => void;
```
RyanMacBern, line 45: Was required before.

src/components/CheckboxInput/CheckboxInput.tsx
```diff
@@ -9,93 +9,72 @@ import './CheckboxInput.scss';
...
   /**
    * Callback function when input is focused
    */
-  onFocus: PropTypes.func,
+  onFocus?: (event: FocusEvent<HTMLInputElement>) => void;
   /**
    * Custom content to be displayed to right of checkbox. Can be any valid node/tree, anchors, etc.
    */
-  label: PropTypes.node.isRequired,
+  label?: string;
```
RyanMacBern, line 53: Was a required node before, now an optional string?

src/components/colors/Colors.stories.tsx
```diff
@@ -1,16 +1,18 @@
 import React from 'react';
 import { withA11y } from '@storybook/addon-a11y';
-import { PALMETTO_COLOR_VALUES } from '../../lib/tokens';
+import { PALMETTO_COLOR_VALUES, ColorEntry, ColorVariation } from '../../lib/tokens';
 import './colors.scss';
 
 const [baseColors, brandColors] = PALMETTO_COLOR_VALUES;
+console.log(baseColors);
+console.log(brandColors);
```
RyanMacBern, line 8: Remove

src/components/CheckboxInput/CheckboxInput.tsx
ericyliu: ah good catch

src/components/CheckboxInput/CheckboxInput.tsx
ericyliu: i can change it back to required, but it was weird that it was required given that we do a check for it when rendering.

it needs to be a string because it's passed into aria-label, which doesnt accept react nodes

src/components/colors/Colors.stories.tsx
ericyliu: woops >.<

PR #64, opened 2020-07-09
47 Select Input

JCarmine:


src/components/Button/Button.jsx
```diff
@@ -28,15 +28,15 @@ const Button = ({
   });
 
   const handleClick = e => {
-    onClick(e);
+    if (onClick) onClick(e);
```
JCarmine, line 31: Just fixing an existing issue with Button. All 3 handlers aren't required, so checking for their presence before trying to fire.

src/components/SelectInput/SelectInput.stories.jsx
```diff
@@ -0,0 +1,104 @@
+import React from 'react';
```
juanfabrega, line 1: @JCarmine when you get a chance check out this PR for a better way to use state in stories. This way the source code is preserved with the true components for users to peruse.

https://github.com/palmetto/palmetto-components/pull/63

src/components/SelectInput/SelectInput.jsx
```diff
@@ -0,0 +1,147 @@
...
+  /**
+   * Options for dropdown list
+   */
+  options: PropTypes.arrayOf(PropTypes.shape({
+    value: PropTypes.string.isRequired,
+    label: PropTypes.string.isRequired,
+  })).isRequired,
+};
+
+export default SelectInput;
```
ericyliu, line 147: so when i was doing the conversion to typescript, i found that the convention should be, if it appears in the defaultProps, it should be required in the propTypes (and storybook/react will mark it as optional)

src/components/SelectInput/SelectInput.jsx
```diff
@@ -0,0 +1,159 @@
...
+   */
+  isDisabled: PropTypes.bool,
+  /**
+   * Determines if input is required or not. (Label will have an asterisk if required)
+   */
+  isRequired: PropTypes.bool,
+  /**
+   * Callback function to call on change event.
+   */
+  onChange: PropTypes.func.isRequired,
```
nathanyoung, line 126: @juanfabrega Does this SelectInput and the TextInput really need to have onChange required?

PR #66, opened 2020-07-09
test (CheckboxInput): add tests

nathanyoung:
adds some basic tests for CheckboxInput component


src/components/CheckboxInput/CheckboxInput.jsx
```diff
@@ -6,6 +6,10 @@ import InputValidationMessage from '../InputValidationMessage/InputValidationMes
 import FormLabel from '../FormLabel/FormLabel';
 import './CheckboxInput.scss';
 
+/**
+ * Use checkbox to allow users to make a range of selections (zero, one or many).
```
nathanyoung, line 10:
```suggestion
 * Used to allow users to make a range of selections (zero, one or many).
```

nathanyoung:
#17

PR #67, opened 2020-07-09
Make ID Required for Controls

JCarmine:


src/components/Button/Button.jsx
```diff
@@ -1,12 +1,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
-import { v4 as uuid } from 'uuid';
 import './Button.scss';
 import { ReactComponent as LoadingIndicator } from '../../images/loading.svg';
 
 const Button = ({
-  id,
```
nathanyoung, line 9: We need allow buttons to be passed an optional ID by the consumer.

src/components/Button/Button.jsx
```diff
@@ -82,10 +76,6 @@ Button.defaultProps = {
 };
 
 Button.propTypes = {
-  /**
-   * A unique identifier for the button
-   */
-  id: PropTypes.string,
```
nathanyoung, line 88: I think we still need this. But we shouldn’t auto gen one.

Issue #50, opened 2020-07-08
Formik Bindings


juanfabrega:
We mean to create formik bindings to all our form components so that they can be used seamlessly with formik.

juanfabrega:
![](https://github.trello.services/images/mini-trello-icon.png) [Formik Bindings](https://trello.com/c/Z5bCbI4M/96-formik-bindings)

juanfabrega:
https://github.com/palmetto/palmetto-components/pull/52

PR #69, opened 2020-07-10
Select input formik adapter

nathanyoung:
attempting to make a Formik adapter for the SelectInput. Getting the following error onChange in the Formik form story example:

- [x] `isMulti` not working in Formik example
- [x] validation doesn't show error beneath SelectInput
- [ ] ~~need to add generateOnChangeHandler to checkbox and text inputs too~~


src/components/Formik/FormikSelectInput/FormikSelectInput.jsx
```diff
@@ -0,0 +1,63 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import SelectInput from '../../SelectInput/SelectInput';
+
+/**
+ * @description Higher order function that provides a closure for the SelectInput onChange handler
+ *  we need to set the values, this is necessary because the Formik Field onChange handler is expecting
+ *  an event object while the SelectInput onChange is only providing the value as the argument
```
juanfabrega, line 8: @nathanyoung I think we generally have a disparity with other components. In TextInput for example, we do pass the event object, which is why this would probably not be needed.

CheckboxInput passes a bool value directly. So we should standardize across the board. For my money I'd rather pass the event directly to the parent, thus always giving the user slightly more responsibility but also a little more flexibility.

nathanyoung:
[Single/Multi Select Input](https://trello.com/c/CQ0txr0k/47-single-multi-select-input)

nathanyoung:
[Formik Bindings](https://trello.com/c/Z5bCbI4M/96-formik-bindings)

nathanyoung:
#50

src/components/Formik/FormikSelectInput/FormikSelectInput.jsx
JCarmine: @nathanyoung @juanfabrega Coming back to this, I also think we should pass the event object. I know this will require some updating on the alchemy side, but I'm happy to see it through.

nathanyoung:
@juanfabrega pushed a fix for displaying validation message, but now the `isMulti` select doesn't work.

juanfabrega:
No worries I will take a look shortly and get it working.

src/components/Formik/FormikSelectInput/FormikSelectInput.jsx
juanfabrega: @JCarmine luckily I think they've all used it with Formik so as long as our formik abstraction is there we should be good.

src/components/Formik/Formik.stories.jsx
```diff
@@ -68,15 +100,58 @@ export const FormikForm = () => {
...
+              <Field
+                label="Flavor no onChange"
+                name="flavor2"
+                id="flavor2"
+                options={flavorOptions}
+                component={FormikSelectInput}
+                // With a custom onChange.
+                // We preserve Formik's convention and relegate state management back to the user.
+                isRequired
+              />
```
nathanyoung: @juanfabrega this Field does not provide an onChange function, but the Formik adapter doesn't wire up the setFieldValue automatically, so it doesn't work.

src/components/Formik/Formik.stories.jsx
```diff
@@ -68,15 +100,55 @@ export const FormikForm = () => {
...
+            </div>
+            <div style={{ marginBottom: '1.25rem' }}>
+              <Field
+                label="Flavor no onChange"
+                name="flavor2"
+                id="flavor2"
+                options={flavorOptions}
+                component={FormikSelectInput}
+                isRequired
+              />
```
nathanyoung, line 136: @juanfabrega this Field does not provide an onChange function, but the Formik adapter doesn't wire up the setFieldValue automatically, so it doesn't work.

src/components/Formik/Formik.stories.jsx
```diff
@@ -68,15 +100,55 @@ export const FormikForm = () => {
...
+              <Field
+                label="Colors"
+                name="colors"
+                id="colors"
+                isMulti
+                options={colorOptions}
+                component={FormikSelectInput}
+                onChange={selectedOption => {
+                  setFieldValue('colors', selectedOption);
+                }}
```
nathanyoung, line 148: If the consumer manually sets the onChange to setFieldValue, then it works. However, I don't think this is our desired experience (should be like TextInput). 

So now I'm thinking we need to go back to the `generateOnChangeHandler()` for SelectInput...

juanfabrega:
@nathanyoung @JCarmine 

In my latest commit I made a few changes to the API as a result of our discussion on Friday. Note that these should probably be discussed to make sure we're all good with the solution. 

First off, we discussed having our APIs be consistent and whenever possible actually bubble up events so that components are as close as possible to a natural react experience. With react-select that unfortunately isn't possible because they have no event to bubble up. I read through this entire thread that discusses this problem and I recommend you do as well:

https://github.com/JedWatson/react-select/issues/1631

I considered different solutions but ultimately landed on a middle ground between the vanilla react select and full on event dispatching. I'm basically generating an object that has some of the same properties as a JS event, but it is NOT an event. At least this way our component APIs are consistent to the user, but they will definitely not get a true event.

If we did want to go that route, an alternate approach worth considering is to render the react-select **and** a _hidden native <select>_ element, which we will use as a target to dispatch events with `new Event` constructor. The main issue with this is that it requires a lot of prop mapping so that our select shares the same properties as the react-select props. It is convoluted, but as close to the real deal as I can think of. Question is whether it is necessary.

With the approach I took, our vanilla component is working now, and the Formik adapter works with their standard API as well. I did have to change our stories to conform to this new API.

Lastly, I also caught an issue with prop naming disparity for errors which was causing the validation message not to show up. We can discuss how we want to name these across the board, but at the moment it seems `error` is the convention which takes either a boolean or the actual error message. I am good with this but if there are concerns about clarity please bring them up.

PR #79, opened 2020-07-13
feat: adding maxlength prop on TextINput

juanfabrega:
NOTE: the babel plugins are to allow async/await usage with jest tests. not used currently but it will be useful when testing events with `userEvent`

### DIFF
--- a/src/components/TextInput/TextInput.jsx
+++ b/src/components/TextInput/TextInput.jsx
@@ -27,6 +27,7 @@ const TextInput = ({
   error,
   isRequired,
   label,
+  maxLength,
   name,
   onBlur,
   onChange,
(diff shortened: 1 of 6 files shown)


src/components/TextInput/TextInput.jsx
```diff
@@ -156,6 +158,11 @@ TextInput.propTypes = {
    * Value for HTML <label> tag
    */
   label: PropTypes.string,
+  /**
+   * The input's 'maxlength' attribute.
+   * NOTE: initializing the input with a value longer than the desired maxlength will not trim this value.
+   */
+  maxLength: PropTypes.string,
```
JCarmine, line 165: It might be more intuitive if this a number.

PR #83, opened 2020-07-13
Select input tests

JCarmine:
Screen Shot 2020-07-13 at 2 20 58 PM


package.json
```diff
@@ -81,12 +81,14 @@
     "identity-obj-proxy": "^3.0.0",
     "jest": "^24.9.0",
     "mini-css-extract-plugin": "^0.9.0",
+    "mutationobserver-shim": "^0.3.7",
```
JCarmine, line 84: Not sure why I had to add this; it had something to do with async/await. Also not sure why placing the import in setupTests.js had no effect, while placing it directly in the component did. Might setupTests.js not be getting initialized correctly?

src/components/SelectInput/SelectInput.test.jsx
```diff
@@ -0,0 +1,283 @@
...
+      console.error = jest.fn(); // eslint-disable-line no-console
+      render(<SelectInput id="testId" onChange={jest.fn()} />);
+      expect(console.error).toHaveBeenCalledTimes(1); // eslint-disable-line no-console
+      expect(console.error.mock.calls[0][0]) // eslint-disable-line no-console
+        .toContain('Failed prop type: The prop `options`');
+    });
+  });
+
+  describe('Callback Handling', () => {
+    // test('it fires onChange callback on change', () => {
```
JCarmine, line 48: I couldn't get this test to work. However, `onChange` is the most important callback. Could someone have a look?

src/components/SelectInput/SelectInput.test.jsx
```diff
@@ -0,0 +1,283 @@
+import React from 'react';
+import Select from 'react-select';
+import '@babel/polyfill';
+import 'mutationobserver-shim';
```
JCarmine, line 4: Both of these imports should be happening in setupTests.js, but putting them there had no effect. Could someone guess why?

package.json
juanfabrega: Hey @JCarmine that's odd but I did have to add some babel plugins to my PR as well to add async/await functionality.

src/components/SelectInput/SelectInput.test.jsx
juanfabrega: @JCarmine I believe these should both be in the .babelrc rather than the setupTests file

juanfabrega: @JCarmine by bad I think you are correct with your initial assessment but I am not sure yet why the `setupTests` file is not loading these dependencies properly. Will continue to investigate.

src/components/SelectInput/SelectInput.test.jsx
```diff
@@ -0,0 +1,284 @@
...
+        );
+        await selectEvent.select(screen.getByLabelText('Ice Cream Flavors'), ['Chocolate', 'Vanilla']);
+        expect(screen.getByTestId('form')).toHaveFormValues({ iceCreamFlavors: ['chocolate', 'vanilla'] });
+
+        await selectEvent.select(screen.getByLabelText('Ice Cream Flavors'), 'Strawberry');
+        expect(screen.getByTestId('form'))
+          .toHaveFormValues({ iceCreamFlavors: ['chocolate', 'vanilla', 'strawberry'] });
+      });
+    });
+  });
```
nathanyoung, line 153: Do we need to be concerned with testing react select itself? If we need to test these scenarios, we should testing our own component.

src/components/SelectInput/SelectInput.test.jsx
```diff
@@ -0,0 +1,284 @@
...
+
+  describe('States', () => {
+    describe('No label, with a placeholder', () => {
+      test('it renders input without a label, and with a placeholder', () => {
```
nathanyoung, line 157: stylistically, do we need to have the nested describe with a single test description? Why not just have the test?

JCarmine: Dunno. You say what you're describing, then describe it.

src/components/SelectInput/SelectInput.test.jsx
juanfabrega: Agreed on this. react-select is technically an implementation detail as far as out users are concerned, so it would make sense to just test our component for these scenarios and let the test fail if for any reason our react-select implementation breaks in the future.

src/components/SelectInput/SelectInput.test.jsx
juanfabrega: I don't feel super strongly on this one way or the other, but I can see how nesting single tests can lead to inconsistency since it's not clear when to nest and when not to. I'd rather use the top-level `describe`  in the test itself and keep the `states` as the parent. The test copy itself seems a bit redundant in these cases.

juanfabrega: To clarify, I would probably do it like this:

```
describe('states', () => {
  test('With a label, and no custom placeholder', () => {
    // test here
  });
});
```

@JCarmine @nathanyoung

src/components/SelectInput/SelectInput.test.jsx
JCarmine: Yeah, I wouldn't have done this had it not been for this link being shared. Was there another reason why this was provided? https://testing-library.com/docs/ecosystem-react-select-event

nathanyoung:
i shared it to reference react-select-event for testing our component,
which uses react-select.

nathanyoung:
@JCarmine The test is failing in the github actions because of propType warnings. You need to pass an onChange function to every <SelectInput /> in since we require that prop.

src/components/SelectInput/SelectInput.test.jsx
```diff
@@ -0,0 +1,358 @@
...
+        );
+        await selectEvent.select(screen.getByLabelText('Ice Cream Flavors'), ['Chocolate', 'Vanilla']);
+        expect(screen.getByTestId('form')).toHaveFormValues({ iceCreamFlavors: ['chocolate', 'vanilla'] });
+
+        await selectEvent.select(screen.getByLabelText('Ice Cream Flavors'), 'Strawberry');
+        expect(screen.getByTestId('form'))
+          .toHaveFormValues({ iceCreamFlavors: ['chocolate', 'vanilla', 'strawberry'] });
+      });
+    });
+  });
```
nathanyoung, line 203:
```suggestion

```

PR #84, opened 2020-07-14
feat: make label required, add aria-labelledby attribute, add hideLabel option

nathanyoung:
Our current label/input implementation makes it easy to break accessibility standards without warning by not including a label for inputs. This PR address that by:

- Makes `label` prop required so that consumers can not break accessibility standards without warnings
- adds `hideLabel` prop to allow for users who want to visually hide the label, but still provide an input label for assistive technology
- Adds `aria-labelledby` attribute to inputs for assistive technology


nathanyoung:
Will update to point to `develop` once `select-input-tests` get's merged.

Issue #87, opened 2020-07-14
setupTest is not applying packages globally


nathanyoung:
make this unecessary https://github.com/palmetto/palmetto-components/blob/develop/src/components/SelectInput/SelectInput.test.jsx#L4

PR #89, opened 2020-07-14
Test Config

JCarmine:
Fixes https://github.com/palmetto/palmetto-components/issues/87

testConfig.js was an obsolete file that I used out of habit from using CRA. Added global dependencies to correct file.


juanfabrega:
Thanks @JCarmine good to go after conflicts resolved.

PR #94, opened 2020-07-14
Code coverage

juanfabrega:


jest.config.js
```diff
@@ -13,4 +13,10 @@ module.exports = {
   },
   coverageDirectory: './coverage/',
   collectCoverage: true,
+  collectCoverageFrom: [
+    '**/*.{js,jsx}',
+    '!**/*.stories.{js,jsx}',
+    '!**/node_modules/**',
+    '!**/vendor/**',
```
nathanyoung, line 20: I might be misunderstanding this, but shouldn't we not collect coverage from these things?

juanfabrega:
The not (!) operator is there for that exact purpose. :)

jest.config.js
juanfabrega: @nathanyoung correct we are excluding these. (!)

nathanyoung: yup, i was misunderstanding 👍 

do we need to add .mdx too?

juanfabrega: @nathanyoung hmm, I think you're correct on that one. Let me confirm.

juanfabrega: @nathanyoung it does not appear that mdx files get picked up by the jest runner at all actually. Which actually makes me wonder if they are getting linted, will confirm but that is a separate PR. This one is good to go.

PR #96, opened 2020-07-15
test (CheckboxInput): increase test coverage and update tests to not be concerned with FormLabel details

nathanyoung:


src/components/CheckboxInput/CheckboxInput.jsx
```diff
@@ -104,7 +104,7 @@ const CheckboxInput = ({
         <input
           aria-invalid={!!error}
           aria-label={label}
-          aria-labelledby={label ? `${id}Label` : null}
+          aria-labelledby={`${id}Label`}
```
nathanyoung, line 107: label is required, so check not necessary

src/components/CheckboxInput/CheckboxInput.test.jsx
```diff
@@ -4,8 +4,19 @@ import {
   fireEvent,
 } from '@testing-library/react';
 import CheckboxInput from './CheckboxInput';
+import FormLabel from '../FormLabel/FormLabel';
+
+jest.mock('../FormLabel/FormLabel');
```
nathanyoung, line 9: Mocking `<FormLabel />`, but making sure we call it with the correct props.

nathanyoung:
@juanfabrega looks like codecov thinks these are new files, so it's counting them agains the coverage diff. How do we update the base coverage number?

Screen Shot 2020-07-14 at 6 08 01 PM

juanfabrega:
@nathan this one is weird because the DIFF is n/a, but honestly I'm pretty sure develop is already at 74% when I merge my last PR that scanned all files. I am working on adding tests to the Formik adapters which should raise the coverage back to 85% hopefully.

juanfabrega:
I would not worry about the the diff on this PR specifically I think it is complaining about the fact that we're off our previous high.

Issue #74, opened 2020-07-10
Radio Button Group Input


nathanyoung:
Radio button component that allows users to select a single option from a list of options.

```jsx
function RadioExample() {
  const [value, setValue] = React.useState("1");
  return (
    <RadioInput
      name="size"
      title="Pick Your Size"
      description="Pick your t-shirt size"
      onChange={e => setValue(e.target.value)}
      options={[
        { value: 'chocolate', label: 'Chocolate' },
        { value: 'strawberry', label: 'Strawberry' },
        { value: 'vanilla', label: 'Vanilla' },
      ]}
      value={value}
    />
  );
}
```

RadioInput Props
| prop  | type | default value |  description |
| ------------- | ------------- | ------------- | ------------- |
| name  | string |  | name attribute that gets applied to all radios |
| className | string |  | additional classNames to add to the group |
| error | bool,string |  | render the radio group in an error state and/or display a validation message |
| onChange | func | undefined | callback function |
| title | node |  | text used as the title for the group of radios |
| description | node |  | help text beneath the title |
| options | array |  | the radio buttons |
| isRequired | bool |  | at least one option must be selected |
| value | string, number|  | value of the radio to be 'checked' |

Radio Props
| prop  | type | default value |  description |
| ------------- | ------------- | ------------- | ------------- |
| name  | string |  | name attribute that gets applied to all radios |
| id | string |  | additional classNames to add to the group |
| value | bool,string |  | value that is returned on form submission |
| isChecked | func |  | if the radio is checked |
| defaultIsChecked | node |  | initially checked if true |
| isDisabled | bool |  | disable selecting the radio |
| error | bool, string |  | render the radio in an error state and/or display a validation message |
| onChange | node |  | callback when input state changes |

nathanyoung:
![](https://github.trello.services/images/mini-trello-icon.png) [Radio Button/Group](https://trello.com/c/3UZwSS0D/110-radio-button-group)

JCarmine:
Noticed that we have both `value` and `defaultValue` as props. Should there be a mode prop for this, or should it be strictly controlled?

All other inputs are strictly controlled, so we might as well follow suit.

nathanyoung:
@JCarmine let's do strictly controlled for now

JCarmine:
Done.

PR #105, opened 2020-07-16
110 Radio Selector

JCarmine:
#74


src/components/RadioGroup/RadioGroup.jsx
```diff
@@ -0,0 +1,166 @@
...
+    label: PropTypes.string.isRequired,
+    isDisabled: PropTypes.bool,
+  })).isRequired,
+  /**
+   * Custom content to be displayed above the RadioGroup.
+   * Both a title and description can be included in the legend.
+   */
+  legend: PropTypes.shape({
+    title: PropTypes.node.isRequired,
+    description: PropTypes.node,
```
nathanyoung, line 38: Instead of a `legend` object prop, can we have separate `title` and `description` props? Makes markup simpler on the consumer end.

src/components/RadioGroup/RadioGroup.jsx
```diff
@@ -0,0 +1,166 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import FormLabel from '../FormLabel/FormLabel';
+import InputValidationMessage from '../InputValidationMessage/InputValidationMessage';
+import './RadioGroup.scss';
+
+/**
+ * Allows users to pick a value from predefined list of options.
```
nathanyoung, line 9:
```suggestion
 * Allows users to pick a single option from predefined list of two or more options.
```

src/components/RadioGroup/RadioGroup.jsx
```diff
@@ -0,0 +1,166 @@
...
+    value: PropTypes.string.isRequired,
+    label: PropTypes.string.isRequired,
+    isDisabled: PropTypes.bool,
+  })).isRequired,
+  /**
+   * Custom content to be displayed above the RadioGroup.
+   * Both a title and description can be included in the legend.
+   */
+  legend: PropTypes.shape({
+    title: PropTypes.node.isRequired,
```
nathanyoung, line 37: `title` shouldn't be required.

if we want to prevent someone from passing a `description` without a `title`, then we should make a "requires other prop" proptype and warn if the consumer sets a description without a title

src/components/RadioGroup/RadioGroup.jsx
nathanyoung: hmmm... reading more of your comments, now I understand better what you are doing with the legend object with a required `title`.

I guess I prefer the latter when it comes to actually using the component
```
<RadioGroupInput
  legend={{ 
    title: 'hello world',
    description: 'the description'
  }}
...
/>

vs

<RadioGroup
  title="hello world"
  description="the description"
  ...
/>
```

If you pass a description without a title, you're still going to get a propType warning.

src/components/FormLabel/FormLabel.jsx
```diff
@@ -4,10 +4,6 @@ import classNames from 'classnames';
 import './FormLabel.scss';
 
 const propTypes = {
-  /**
```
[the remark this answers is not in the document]
juanfabrega: @JCarmine Interesting. Not sure yet if I am for or against. I have been doing alphabetical order on mine, which makes it easy to find what I'm lookin for. Let's discuss today.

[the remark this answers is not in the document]
JCarmine: Ah, I've been going (loosely) by "most important". Would be a good pattern to nail down.

src/components/RadioGroup/RadioGroup.jsx
```diff
@@ -0,0 +1,166 @@
...
+                  id={option.id}
+                  type="radio"
+                  name={name}
+                  className="radioInput"
+                  value={option.value}
+                  checked={selectedOption === option.value}
+                  onChange={onChange}
+                  onFocus={handleFocus}
+                  onBlur={handleBlur}
+                  disabled={isDisabled || option.disabled}
```
juanfabrega, line 150: @JCarmine should this be `option.isDisabled` I believe that is the shape noted in the proptypes, unless I am missing something.

JCarmine:
> @JCarmine this looks great here are my only comments so far:
> 
> 1. Can we take a slightly more modular approach here and create a `RadioInput` that can then be used as a component in the `RadioGroup`. This is similar to how we have used the `FormLabel` and the InputValidationMessage` I think in general this helps keep things more tightly encapsulated, re-usable, and more easily testable. I realize a single radio input isn't really a use case, so this is more of just a code structure request rather than user-facing request.
> 2. Can we include a prop that governs whether the RadioInputs in a group are rendered horizontally or vertically? (`inline: PropTypes.bool`) Perhaps this prop flows down to the RadioInput div itself adding an `inline` display or something like that implementation up to you.

Re: #1: Yes, I can break RadioInput into a separate component. It's folder will be placed inside RadioGroup, until there's a use-case for it being a stand-alone component.
Re: #2: According to the issue and original specs, this would be scope-creep. However, I'm happy to do it if @nathanyoung thinks it worth doing at this time.

src/components/RadioGroup/RadioGroup.jsx
juanfabrega: Seems that because that specific part of the shape is not required, it is not throwing a proptypes error, but there is a mismatch.

nathanyoung:
> 2: According to the issue and original specs, this would be scope-creep. However, I'm happy to do it if @nathanyoung thinks it worth doing at this time.

Let's wait on this until we have a better handle on how we're going to manage horizontal and vertical flow.

src/components/FormLabel/FormLabel.jsx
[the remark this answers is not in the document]
juanfabrega: @JCarmine follow the vote in the ux channel. 

ALPHA required
ALPHA non-required

src/components/FormLabel/FormLabel.module.scss
```diff
@@ -6,11 +6,28 @@
   font-size: var(--size-font-sm);
```
juanfabrega, line 6: @JCarmine maybe this is best served for another PR, but I'm thinking we should have this follow our pattern of component-specific variables similarly to `Button` etc.. Let me know what you think.

src/components/RadioGroup/RadioGroup.module.scss
```diff
@@ -0,0 +1,29 @@
+.radio-group {
```
juanfabrega, line 1: Same comment as above, let's decide if we want to add the component-level variables here or in a separate PR.

src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,157 @@
...
+import classNames from 'classnames';
+import RadioInput from './RadioInput/RadioInput';
+import InputValidationMessage from '../InputValidationMessage/InputValidationMessage';
+import styles from './RadioGroup.module.scss';
+
+/**
+ * Allows users to pick a single option from a predefined list of two or more options.
+ * Ideally, it should be used when there are less than 5 options,
+ * otherwise you should consider using SelectInput instead.
+ */
```
JCarmine, line 12: Actually, this seems like documentation, and since Storybook is so good at managing this, I'm wondering if we should just remove it in the files.

This would have the added benefit of ensuring that the two sources of documentation don't get out-of-step.

src/components/Formik/FormikRadioGroup/FormikRadioGroup.test.jsx
```diff
@@ -0,0 +1,185 @@
+import React from 'react';
+import {
+  render,
+  fireEvent,
+  screen,
+  waitFor,
+} from '@testing-library/react';
+import selectEvent from 'react-select-event';
+import { Formik, Form, Field } from 'formik';
+import FormikSelectInput from './FormikSelectInput';
```
nathanyoung, line 10: Wrong component

JCarmine: @nathanyoung Yeah, I hadn't started seriously adding tests at that point. Tests will be up for review tomorrow.

src/components/FormLabel/FormLabel.module.scss
nathanyoung: Updated to use form variables, and added radio specific ones too.

src/components/RadioGroup/RadioGroup.module.scss
nathanyoung: Updated to use form variables, and added radio specific ones too.

nathanyoung:
@JCarmine made some styling tweaks and addressed the visual regressions picked up by chromatic.

.vscode/settings.json
```diff
@@ -1,5 +1,6 @@
 {
   "cSpell.words": [
+    "formik",
```
JCarmine, line 3: I'm adding these words to the vscode workspace so that the IDE quits nagging when someone names a package something silly, like "formik". Is this beneficial?

src/components/Formik/Formik.stories.jsx
```diff
@@ -158,7 +189,9 @@ export const FormikForm = () => {
               />
             </div>
             <Button type="submit" isLoading={isSubmitting}>Submit</Button>
-            <pre>{JSON.stringify(values, isSubmitting, null, 2)}</pre>
+            <pre>
+              {JSON.stringify(formatOutput(values, isSubmitting), undefined, 2)}
```
JCarmine, line 193: Screen Shot 2020-07-29 at 11 41 30 AM

src/components/RadioGroup/RadioGroup.stories.mdx
```diff
@@ -0,0 +1,338 @@
+import RadioGroup from './RadioGroup';
```
JCarmine, line 1: Proofread, proofread, proofread!

.vscode/settings.json
juanfabrega: @JCarmine generally I'd recommend keeping any editor-specific settings/files out of the repo, and making these changes only locally while they stay in gitignore. I get it's probably annoying but it feels inherently personal to the user.

src/components/FormLabel/FormLabel.tsx
```diff
@@ -16,34 +16,58 @@ interface Props {
...
   /**
    * Mark the label has invalid
    */
   hasError?: boolean;
-   /**
+  /**
    * Mark the label has disabled
    */
-  isDisabled?: boolean;
+  isDisabled?: boolean,
```
juanfabrega, line 30: @JCarmine in an inteface these should end in semicolons, not commas, unless I'm missing something.

src/components/FormLabel/FormLabel.tsx
```diff
@@ -16,34 +16,58 @@ interface Props {
    * Custom class to pass to label element.
    */
   className?: string;
+  /**
+   * Display label inline with surrounding elements
+   */
+  displayInline?: boolean,
```
juanfabrega, line 22: Semicolon not comma here.

src/components/RadioGroup/RadioGroup.stories.mdx
```diff
@@ -0,0 +1,338 @@
...
+
+<Meta
+  title="Components/Form Inputs/RadioGroup"
+  component={RadioGroup}
+  decorators={[withA11y]}
+/>
+
+# RadioGroup
+
+Use a RadioGroup when a user is required to select one of five or less options. It is ideal for this scenario because the options are displayed without having to interact. If there are more than five options, use the [SelectInput](?path=/docs/components-form-inputs-selectinput--all) instead.
```
juanfabrega, line 15: five or **fewer**

:D

src/components/RadioGroup/RadioGroup.stories.mdx
```diff
@@ -0,0 +1,338 @@
...
+## Default
+
+All that is required to render a basic version of the RadioGroup is the group's `name`, an onchange event handler passed to the `onChange` prop, and a list of options passed to the `options` prop. Each option must include a unique `id`, a `value`, and a `name`.
+
+<Preview isExpanded>
+  <Story name="Default">
+    {() => {
+      const [value, setValue] = useState(null);
+      const options = [{
+        id: 'chocolate',
```
juanfabrega, line 30: The copy above says we need `id` `value` and `name` but these don't have `name` properties in the options

src/components/FormLabel/FormLabel.tsx
JCarmine: Actually, since we are no longer writing JS, we can do what ever we want! Comma, semi, or nothing, anything goes! What's with programming languages mandating a specific syntax to function properly anyways?

But in all seriousness, this rule wasn't being enforced out of the box for whatever reason, so we had inconsistencies in other files as well. Added a rule for it to `.eslintrc`. Feel free to tweak.

src/components/RadioGroup/RadioGroup.stories.mdx
JCarmine: Thanks for the careful proofing; appreciate it!

.eslintrc
```diff
@@ -55,7 +55,17 @@
...
+    "@typescript-eslint/member-delimiter-style": ["error", {
+      "multiline": {
+        "delimiter": "semi",
+        "requireLast": true
+      },
+      "singleline": {
+        "delimiter": "semi",
+        "requireLast": true
+      }
+    }]
```
JCarmine, line 68: Please modify as you see fit.

src/components/FormLabel/FormLabel.tsx
juanfabrega: Perfect, thanks! @JCarmine

PR #107, opened 2020-07-17
Button size

nathanyoung:
do not merge, still need to publish updated design tokens package


juanfabrega:
@nathanyoung this PR requires an update of the palmetto-design-tokens dependency, right?

nathanyoung:
@juanfabrega correct, it uses border-radius tokens that don't exist in the design tokens package.

juanfabrega:
@nathanyoung closing this since it is included here https://github.com/palmetto/palmetto-components/pull/120

PR #120, opened 2020-07-21
Css modules

juanfabrega:
WIP for the CSS modules proof of concept. A few things that still need some attention.

~~1. react-select is not inheriting styles since they are scoped to a module. Need to confirm best approach here but likely a global stylesheet for these specifically.~~

~~2. Utility classes declared as :global currently. necessary to avoid importing them in every single module (which massively inflates bundle size).~~

@nathanyoung I merged in your button-sizes branch in here to see if we can perhaps pull it in simultaneously.


src/components/Spinner/Spinner.module.scss
```diff
@@ -0,0 +1,24 @@
+.spinner {
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
+
+  & svg {
```
RyanMacBern, line 7: ampersand isn't necessary

src/components/Spinner/Spinner.module.scss
```diff
@@ -0,0 +1,24 @@
...
+    display: inherit;
+    -webkit-animation: spin 1s infinite linear;
+    animation: spin 1s infinite linear;
```
RyanMacBern, line 10: Are we not using the Autoprefixer PostCSS plugin for palmetto-components?

src/components/SelectInput/SelectInput.module.scss
```diff
@@ -0,0 +1,72 @@
+@import '../../styles/variables.scss';
+
+:global {
+  .selectInputWrapper {
+    .selectInput {
+      .selectInput__control {
```
RyanMacBern, line 6: There's no use in having this a CSS module if the entire file is wrapped in :global. I would try one of these options:

```
.otherClassName {
  composes: className from "./style.css";
}
```

```
.otherClassName {
  composes: globalClassName from global;
}
```

src/components/Spinner/Spinner.module.scss
juanfabrega: It is not configured yet, but that is the plan!

src/components/SelectInput/SelectInput.module.scss
juanfabrega: Yup, agreed, this one is still in progress, and not working properly. Adding the global was the last step I did last night to get it working properly but it still isn't properly passing the styles down to our react-select. I plan to not have it be a module altogether and just import the styles.

src/components/Button/Button.module.scss
```diff
@@ -0,0 +1,73 @@
...
+
+  // medium (default)
+  --button-size-md-border-radius: 8px; // var(--size-border-radius-md);
+  --button-size-md-font-size: var(--size-font-md);
+  --button-size-md-padding: var(--size-spacing-sm) var(--size-spacing-md);
+
+  // large
+  --button-size-lg-border-radius: 8px; // var(--size-border-radius-md);
+  --button-size-lg-font-size: var(--size-font-lg);
+  --button-size-lg-padding: var(--size-spacing-md) var(--size-spacing-lg);
```
nathanyoung, line 22: @juanfabrega we talked about this offline, but can't remember what we decided. do you think we should move these to `/styles/variables.css`?

nathanyoung:
Color stories are also broken. You can see that in chromatic.

https://www.chromatic.com/snapshot?appId=5ed9214b642dc10022b50a2d&id=5f176ee7f82a0800222fdab6

src/components/FormLabel/FormLabel.scss
```diff
@@ -1,21 +0,0 @@
...
-  }
-
-  +input[type=text],
-  +input[type=tel],
-  +input[type=password],
-  +input[type=search],
-  +input[type=url],
-  +.selectInput {
-    margin-top: $size-spacing-xs;
-  }
```
RyanMacBern, line 20: This isn't needed anymore or is it just moved somewhere else that I'm not seeing?

JCarmine: @juanfabrega This spacing is defiantly absent; border-radius is missing as well.

Screen Shot 2020-07-22 at 9 45 17 AM

src/components/Button/Button.module.scss
juanfabrega: @nathanyoung I am currently working on this re-structure.

src/components/FormLabel/FormLabel.scss
juanfabrega: @RyanMacBern @JCarmine I re-structured these so that the Inputs actually pass a className prop to the label. That way the label is not necessarily concerned with its context and instead the context governs the label.

Can y'all do a `yarn` to make sure you've got the latest tokens package locally? The inputs now pass a utility class but the previous version of tokens did not have these classes. Here is where that class is passed in the `TextInput` component:

```
const labelProps = {
    isFieldRequired: isRequired,
    inputId: id,
    labelText: label,
    hasError: !!error,
    className: 'm-bottom-xs',
  };
```

Again I think this is a better architectural pattern since we don't have non-explicit pollution into our input styles from an external component.

nathanyoung: Remember, Chromatic picks up visual changes, so a change like this would be caught (and would have to be approved). You can always view the PR storybook that's created by the CI. 

Screen Shot 2020-07-22 at 10 23 30 AM

src/components/CheckboxInput/CheckboxInput.tsx
```diff
@@ -21,7 +21,7 @@ interface Props {
   /**
    * Custom content to be displayed to right of checkbox.
    */
-  label: string;
+  label: React.ReactNode;
```
RyanMacBern, line 24: This doesn't cause an issue when a non-string is used for aria-label?

juanfabrega: @RyanMacBern I think you're correct here, although we have provided the ability for the team to pass nodes so that they may add things like anchor tags (which are very typical in something like a terms checkbox). Perhaps the answer here is to allow passing of these attributes separately?

src/components/Button/Button.tsx
```diff
@@ -1,7 +1,8 @@
 import React, { FC } from 'react';
 import classNames from 'classnames';
+import styles from './Button.module.scss';
```
JCarmine, line 3: Imports are another area where it would be good to standardize how they're organized. I generally do: 3rd party imports, then internal component imports, then css/scss. So this block would be:

```
import React from 'react';
import PropTypes from 'prop-types';
import Spinner from '../Spinner/Spinner';
import './Button.scss';	import styles from './Button.module.scss';
```
Or, we could even add spaces between the groups to distinguish them even more. I generally only add a space between the js and css; so: 

```
import React from 'react';
import PropTypes from 'prop-types';
import Spinner from '../Spinner/Spinner';

import './Button.scss';	import styles from './Button.module.scss';
```

Or, we could add a space between each:

```
import React from 'react';
import PropTypes from 'prop-types';

import Spinner from '../Spinner/Spinner';

import './Button.scss';	import styles from './Button.module.scss';
```

src/components/Button/Button.tsx
```diff
@@ -58,17 +63,24 @@ const Button: FC<Props> = ({
...
   onFocus,
   onBlur,
 }) => {
   const disabled = isLoading || isDisabled;
 
-  const buttonClasses = classNames('Palmetto-Button', className, {
-    loading: isLoading,
-    fullWidth,
-  });
+  const buttonClasses = classNames(
```
JCarmine, line 74: I prefer automatically wrapping arguments to classNames to the next line, as is done here. Is this something we should standardize?

src/components/CheckboxInput/CheckboxInput.module.scss
```diff
@@ -0,0 +1,17 @@
+@import '../../styles/variables/tokens.scss';
```
JCarmine, line 1: This import shouldn't be necessary since variables are applied to the :root. Commenting out this import has no impact on the component.

src/components/FormLabel/FormLabel.module.scss
```diff
@@ -0,0 +1,12 @@
+@import '../../styles/variables/tokens.scss';
```
JCarmine, line 1: This import shouldn't be necessary since variables are applied to the :root. Commenting out this import has no impact on the component.

src/components/SelectInput/ReactSelect.scss
```diff
@@ -0,0 +1,72 @@
+@import '../../styles/variables/forms.scss';
```
JCarmine, line 1: This import shouldn't be necessary since variables are applied to the :root. Commenting out this import has no impact on the component.

src/components/SelectInput/SelectInput.jsx
```diff
@@ -154,18 +152,19 @@ const SelectInput = ({
     inputId: id,
     labelText: label,
     hasError: !!error,
+    className: 'm-bottom-xs',
   };
 
   return (
     <>
-      <div className={classNames('Palmetto-SelectInput', className, { disabled: isDisabled })}>
+      <div className={classNames('selectInputWrapper', className, { disabled: isDisabled })}>
```
JCarmine, line 160: Should the standard be that each argument is on its own line?

src/components/SelectInput/SelectInput.module.scss
```diff
@@ -0,0 +1 @@
+/* Placeholder SCSS file for future updates */
```
JCarmine, line 1: There should probably only be one scss file in a dir. What is the reason for having two?

src/components/Spinner/Spinner.tsx
```diff
@@ -0,0 +1,74 @@
...
+    if (size === 'sm') pixels = '12';
+    if (size === 'md') pixels = '16';
+    if (size === 'lg') pixels = '24';
+
+    return pixels;
+  };
+
+  return (
+    <span className={classes}>
+      <svg
```
JCarmine, line 40: It would be nice if we could store this svg in the images folder, and import using svgr. Could we set the width and height of the svg to 100% and wrap it in a parent div that has its dimensions constrained? Also, might it be possible to set the `color` on the container and have it cascade into the svg?

src/components/TextInput/TextInput.jsx
```diff
@@ -3,9 +3,10 @@ import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import Cleave from 'cleave.js/react';
 import * as InputMasks from './TextInputMasks';
-import './TextInput.scss';
+import styles from './TextInput.module.scss';
 import FormLabel from '../FormLabel/FormLabel';
 import InputValidationMessage from '../InputValidationMessage/InputValidationMessage';
+import '../../styles/utilities.scss';
```
JCarmine, line 9: Once again, this seems really out of order. See comment on Button.jsx.

src/components/colors/Colors.module.scss
```diff
@@ -1,16 +1,18 @@
-.colors__color-block__item {
+@import '../../styles/variables/tokens.scss';
```
JCarmine, line 1: This import shouldn't be necessary since variables are applied to the :root. Commenting out this import has no impact on the component.

src/styles/globals/form-controls.scss
```diff
@@ -0,0 +1,47 @@
+@import '../variables/forms.scss';
```
JCarmine, line 1: This import shouldn't be necessary since variables are applied to the :root. Commenting out this import has no impact on the component.

src/styles/variables/buttons.scss
```diff
@@ -0,0 +1,24 @@
+@import './tokens.scss';
+@import './forms.scss';
+
+:root {
```
JCarmine, line 4: Placing this in styles/variables implies that these vars will be used elsewhere. What is the use case where this would happen?

src/styles/variables/index.scss
```diff
@@ -0,0 +1,3 @@
+@import './tokens.scss';
```
JCarmine, line 1: I believe the build is currently creating a css file for each of these individually.

src/styles/variables/tokens.scss
```diff
@@ -0,0 +1,2 @@
+@import '~@palmetto/palmetto-design-tokens/build/css/variables-color.css';
```
JCarmine, line 1: Should token be a css or scss file? In this case, seems like it should be css since it contains exclusively css variables. In addition, it doesn't do any Sassy things.

src/components/TextInput/TextInput.jsx
RyanMacBern: https://eslint.org/docs/rules/sort-imports could be used.

Here's the order rule I'm using for the home app:

`"import/order": [
      "error",
      { "groups": ["object", "builtin", "external", "internal", "parent", "sibling", "index"] }
    ],`

src/components/Spinner/Spinner.tsx
```diff
@@ -0,0 +1,74 @@
...
+            fill={`${PALMETTO_BRAND_COLOR_VALUES[color].base.value}`}
+            d="M8,16c-1.199,0-2.352-0.259-3.428-0.77l0.857-1.807C6.235,13.806,7.1,14,8,14c3.309,0,6-2.691,6-6 s-2.691-6-6-6S2,4.691,2,8c0,0.901,0.194,1.766,0.578,2.572l-1.806,0.859C0.26,10.354,0,9.2,0,8c0-4.411,3.589-8,8-8s8,3.589,8,8 S12.411,16,8,16z" /* eslint-disable-line max-len */
+          />
+        </g>
+      </svg>
+    </span>
+  );
+};
+
+Spinner.propTypes = {
```
JCarmine, line 59: Defining the types in interface, and then defining again in propTypes just doesn't seem right, and I don't feel it's a workable pattern. Ideally, we want the types declared once, at the top of the file.

Perhaps, we can us a package like the following, to auto generated the propTypes based on the TypeScript types?
https://github.com/milesj/babel-plugin-typescript-to-proptypes

src/components/FormLabel/FormLabel.module.scss
```diff
@@ -0,0 +1,12 @@
+@import '../../styles/variables.scss';
+
+.label {
+  font-size: var(--size-font-sm);
+  color: var(--color-font-base);
+  display: block;
```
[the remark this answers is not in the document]
juanfabrega: @JCarmine I dig it! Can we create an issue to add this to our stylelint config for enforcing it automatically? We can use this: https://www.npmjs.com/package/stylelint-order or another alternative.

src/components/SelectInput/SelectInput.jsx
```diff
@@ -4,7 +4,7 @@ import classNames from 'classnames';
 import Select from 'react-select';
 import FormLabel from '../FormLabel/FormLabel';
 import InputValidationMessage from '../InputValidationMessage/InputValidationMessage';
-import './SelectInput.scss';
+import './ReactSelect.scss';
```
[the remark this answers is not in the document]
juanfabrega: Hey @JCarmine the reason I named it ReactSelect is because these styles are applied to the actual ReactSelect component which happens to be housed in our own SelectInput. The ReactSelect also has to be a global to work properly with the prefixing that react-select uses. Perhaps the best way is to move this to the /globals folder, but to be clear I would still like us to have a module for SelectInput so that any additional styling that doesn't belong to react-select can be done there in the future.

rollup.config.js
```diff
@@ -41,19 +41,33 @@ export default {
...
       targets: [
         {
-          src: 'src/styles/variables.scss',
+          src: 'src/styles/variables/index.scss',
           dest: 'dist',
-          rename: '/scss/variables.scss',
+          rename: '/css/variables.css',
+        },
+        {
+          src: 'src/styles/variables/buttons.scss',
```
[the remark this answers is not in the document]
juanfabrega: Yes we are making these variables available to the consumer. The main use case is to allow them to create consistency even when they are not using our components. E.G: If one of our components falls short of their feature needs and they reach for something custom, they could potentially maintain consistency with the rest of the application. I expect the need here to diminish over time, but given that we're still low on features. This is comparable to exposing our tokens (which we do as well) but goes one step further by providing actual component variables.

src/components/colors/Colors.module.scss
juanfabrega: Right you are. I was still in the sass compile-time mindset.

src/styles/variables/buttons.scss
juanfabrega: Answered this elsewhere

src/styles/variables/tokens.scss
juanfabrega: yes I intend to actually rename all these because we're not using sass features anywhere, however, the storybook webpack config complains at the moment when we do this. Needs some tweaking. Perhaps the next item to tackle as a group is improving dev/build process a bit.

src/components/Spinner/Spinner.tsx
juanfabrega: @JCarmine agree this isn't ideal. It is a compromise we came to on a temporary basis due to docgen issues rather than actual issues with the components themselves. The plan is definitely to tackle this further down the line. That babel plugin looks promising.

nathanyoung:
[CSS Architecture](https://trello.com/c/5l1ILieN/116-css-architecture)

src/components/Button/Button.jsx
```diff
@@ -0,0 +1,141 @@
+import React from 'react';
```
JCarmine, line 1: We somehow ended up with both jsx and tsx versions of this Component. Any improvements to Button should be salvaged from this file, and it should be trashed.

src/components/Button/Button.tsx
JCarmine: We really should settle on a pattern for this. Until then, can we have the style imports come after the JS?

I've created an issue to standardize this across the repo: #127

src/components/CheckboxInput/CheckboxInput.module.scss
```diff
@@ -0,0 +1,16 @@
+
```
JCarmine, line 1: Remove extra line from top.

src/components/SelectInput/SelectInput.jsx
```diff
@@ -4,7 +4,7 @@ import classNames from 'classnames';
 import Select from 'react-select';
 import FormLabel from '../FormLabel/FormLabel';
 import InputValidationMessage from '../InputValidationMessage/InputValidationMessage';
-import './SelectInput.scss';
+import '../../styles/globals/react-select.scss';
```
JCarmine, line 7: After further consideration, I don't believe this file belongs in the same dir with utilities. Regardless of its styles being wrapped in `:global`, they are local, and very specific to the component. Since we will never import react-select into another component, this file should probably just be the sole scss file in the component's dir.

src/components/SelectInput/SelectInput.jsx
JCarmine: Let's start putting each argument on a new line. It looks more clean, and improves consistancy.

src/components/TextInput/TextInput.jsx
```diff
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import Cleave from 'cleave.js/react';
 import * as InputMasks from './TextInputMasks';
-import './TextInput.scss';
+import styles from './TextInput.module.scss';
```
JCarmine, line 6: Can we make styles the last import here?

src/components/colors/Colors.module.scss
```diff
@@ -1,16 +1,16 @@
-.colors__color-block__item {
+.colorBlock {
   padding: 1rem;
   color: white;
```
JCarmine, line 3: Should we enforce that all colors come from our design tokens, even for colors as basic as black and white?

src/styles/globals/react-select.scss
```diff
@@ -0,0 +1,70 @@
+:global {
```
JCarmine, line 1: See other comment. I don't think this file belongs here.

src/components/colors/Colors.module.scss
juanfabrega: @JCarmine yes!

src/components/Button/Button.jsx
nathanyoung: agree, remove

nathanyoung:
See comments on visual changes:
https://www.chromatic.com/pullrequest?appId=5ed9214b642dc10022b50a2d&number=120&view=activity

src/components/colors/Colors.stories.tsx
```diff
@@ -31,12 +31,13 @@ const renderColorPalette = (colorEntry: [string, ColorEntry], index: number) =>
   return (
     <div key={index}>
       <h2 style={{ marginTop: '0' }}>{colorName}</h2>
-      {Object.entries(colorVariations).map((colorVariationEntry, index) => {
+      {Object.entries(colorVariations).map((colorVariationEntry, i) => {
```
JCarmine, line 34: Not a blocker, but I'm definitely a fan of being more verbose and just writing out "index".

nathanyoung: In this case there is already an `index` defined on line 32, so lint was complaining.

src/components/SelectInput/SelectInput.scss
```diff
@@ -1,3 +1,5 @@
+/* The following styles required the :global tag since they need to be applied to
+react-select elements for which we do not own the markup */
 :global {
   .selectInputWrapper {
```
RyanMacBern: :global means nothing outside of a module.

I would make this a module and reference the globals like so:
```
.selectInputWrapper {
  :global {
    .reactStyle {
       ...
    }
  }
}
```

juanfabrega: @RyanMacBern due to the way react-select is styled, where we have to override classes on elements we don't render ourselves, having these inside of a module, with :global or not, broke that implemenation. We can perhaps continue experimenting as to why that happens but at the moment this seems to be the best approach.

RyanMacBern: I think the problem is likely because you have a non-global (selectInputWrapper) mixed with the globals. My example should fix that.

RyanMacBern:
I'd recommend adding https://www.npmjs.com/package/eslint-plugin-css-modules. This will make sure you have no unused classes.

Issue #123, opened 2020-07-23
Remove need for duplicate Interface + PropType declarations to generate storybook docs.


juanfabrega:

nathanyoung:
![](https://github.trello.services/images/mini-trello-icon.png) [Babel plugin for generating proptypes](https://trello.com/c/P8q8slu8/138-babel-plugin-for-generating-proptypes)

nathanyoung:
https://storybook.js.org/docs/configurations/typescript-config/

juanfabrega:
This is addressed in the bundling changes that now generate proptypes from typescript interfaces.

PR #129, opened 2020-07-23
Lint TypeScript files

JCarmine:
Linter is running on tsx files. However, we have two outstanding issues:

1) The linter has an issue when trying to import tsx files. Solution might be here: https://www.npmjs.com/package/eslint-import-resolver-typescript
2) The linter is complaining that we are not setting defaults for non-required props. Perhaps we can just silence this error until we work out how auto-generate PropTypes based on TypeScript types. This will be address in this issue: https://github.com/palmetto/palmetto-components/issues/123
Possible solution: https://github.com/milesj/babel-plugin-typescript-to-proptypes

Screen Shot 2020-07-23 at 12 24 10 PM
Screen Shot 2020-07-23 at 12 24 27 PM


RyanMacBern:
Try adding:
```
"settings": {
    "import/resolver": {
      "node": {
        "extensions": [".js", ".jsx", ".ts", ".tsx"]
      }
    }
  }
```

mtblanton:
Ryan's suggestion for the eslint config worked well! The errors around imports are gone.

I've also fixed issues that weren't related to default props.

ericyliu:
i tried using https://github.com/milesj/babel-plugin-typescript-to-proptypes, but thats the library that doesnt generate proptypes properly if the types arent defined in the same file as the component. Besides, that wont remove the error about default props. I think we can remove that error for now, because otherwise we'd just be setting a lot of things to undefined right?

JCarmine:
> i tried using https://github.com/milesj/babel-plugin-typescript-to-proptypes, but thats the library that doesnt generate proptypes properly if the types arent defined in the same file as the component. Besides, that wont remove the error about default props. I think we can remove that error for now, because otherwise we'd just be setting a lot of things to undefined right?

I don't think removing the error is the solution at this point. Seeing as how TypeScript isn't required to have a successful project, I don't see why we should be making concessions for it. Declaring the types in TS, and then repeating ourself in PropTypes, and then hacking the defaultProps (so that we don't have to ignore a logical linter rule) to make it all work just doesn't seem very developer like. The solution has to be either generate one from the other, or drop one or the other, IMHO.

mtblanton:
I set the rules to ignore default props without reading the last comment. We can take that out before merging if we want to keep the rule around.

I personally prefer using default parameters over defaultProps, but there's not a huge functional difference between them.

We can keep playing with proptypes generators, there's some value in having both (especially in a library meant to be consumed by other projects). I found https://github.com/danReynolds/typescript-proptypes-generator that may help with things since the babel plugin isn't quite there yet.

PR #130, opened 2020-07-24
Add Linting for Styles

JCarmine:


src/components/Button/Button.scss
```diff
@@ -2,28 +2,28 @@
 
 .Palmetto-Button {
   position: relative;
-  padding: $size-spacing-sm $size-spacing-md;
+  transition-duration: 0.2s;
```
JCarmine, line 5: This looks scary, but it's actually just stylelint auto-fixing the property order to match the concentric css model: https://rhodesmill.org/brandon/2011/concentric-css/

JCarmine: `npx stylelint "**/*.{scss,sass}" --fix`

PR #139, opened 2020-07-28
SelectInput Typescript

ericyliu:


src/components/SelectInput/SelectInput.tsx
```diff
@@ -188,7 +173,82 @@ const SelectInput = ({
   );
 };
 
-SelectInput.propTypes = propTypes;
-SelectInput.defaultProps = defaultProps;
+SelectInput.propTypes = {
```
JCarmine, line 176: I think we should remove the comments/descriptions from the PropTypes. This will make it more clear where these are being sourced from to render the storybook docs.

Also, we decided a while back that we want to define the props before the function. So in the days of TS, I believe a component should generally look like:

```
interface Props {
 /**
    * Description
    */
  someProp?: boolean,
}

const propTypes = {
  someProp: PropTypes.string
};

const SomeComponent: FC<Props> = ({
  someProp = false
}) => {
...
};

SomeComponent.propTypes = propTypes;

export default SomeComponent;
```

src/components/SelectInput/SelectInput.tsx
```diff
@@ -12,119 +12,104 @@ import './SelectInput.scss'; // Not a module because it requires :global styles
...
-  name: '',
-  onFocus: undefined,
-  onBlur: undefined,
-  autoFocus: false,
-  isMulti: false,
-  value: undefined,
-};
-
-const SelectInput = ({
+const SelectInput: React.FC<Props> = ({
```
JCarmine, line 93: For Button, we import `FC` at the top, and then call it as a standalone. I don't have a strong opinion either way on how it's done, but it should be one or the other. Perhaps either @nathanyoung or @juanfabrega has an opinion on this?

```
// Button.tsx
import React, { FC } from 'react';

const Button: FC<Props> = ({
```

juanfabrega: I prefer a single React import statement personally, But no strong opinion either, let's just make all of them consistent. @ericyliu choose your favorite and enforce it on all our tsx files in this PR.

src/components/SelectInput/SelectInput.tsx
ericyliu: oh good point, i'll make both changes

src/components/SelectInput/SelectInput.tsx
ericyliu: yeah true, im ok with either way

ericyliu: i'll change everything to a single React import

src/components/CheckboxInput/CheckboxInput.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { ChangeEvent, FocusEvent, FC } from 'react';
+import React, { ChangeEvent, FocusEvent } from 'react';
```
JCarmine, line 1: Following the single import at the top convention, would we just have the React import at the top and then `React.ChangeEvent` and `React.FocusEvent` in the document?

JCarmine: I think it might make what the component is depending on more obvious if we just do all of the imports at the top.

ericyliu: seems like juan is in agreement so ill change it back

PR #140, opened 2020-07-28
Enforce kebab-case for both class and id selectors.

JCarmine:
https://stylelint.io/user-guide/rules/regex


src/components/CheckboxInput/CheckboxInput.module.scss
```diff
@@ -2,8 +2,7 @@
   display: flex;
   align-items: center;
 
-  &.isDisabled {
-    color: var(--color-font-grey-light);
```
JCarmine, line 6: Prop is now passed to FormLabel to grey it out.

JCarmine:
Disabled state added to FormLabel; impacts all form inputs.

Screen Shot 2020-07-28 at 11 09 44 AM
Screen Shot 2020-07-28 at 11 09 59 AM
Screen Shot 2020-07-28 at 11 10 34 AM

Issue #138, opened 2020-07-27
Change typescript compile target to es6


ericyliu:
change from `esnext` to `es6`

PR #148, opened 2020-07-31
Consistent bundling

juanfabrega:
#138 This PR is meant to bring webpack bundling to production builds and create a config that is _relatively_ consistent between our build and the storybook dev/build config. 

FEATURES
* type checking now runs on builds rather than after builds.
* Type checking now runs on storybook builds.
* Webpack allows us to more easily minify and tree-shake. Reducing our overall bundle size for our distributable.


.storybook/main.js
```diff
@@ -1,79 +1,19 @@
...
-    options: assetRule.options || assetRule.query
-  };
-  // Merge our rule with existing assetLoader rules
-  config.module.rules.unshift({
-    test: /\.svg$/,
-    use: ["@svgr/webpack", assetLoader]
-  });
-  return config;
-}
+// const svgRules = config => {
```
mtblanton, line 4: There's a few instances of commented code in the changes. Are we keeping those around for a reason?

juanfabrega: @mtblanton I was wanting to keep this config as I think we we will be needing it soon(ish) but I much prefer to delete it and start clean once we get to that point where we need to process svgs in our builds.

PR #151, opened 2020-08-04
All Storybook to MDX

JCarmine:


src/components/CheckboxInput/CheckboxInput.stories.mdx
```diff
@@ -0,0 +1,193 @@
...
+
+Use a CheckboxInput to allow users to make a range of selections (zero, one or many).
+
+## Props
+
+<Props of={CheckboxInput} />
+
+## Default
+
+All that is required to render a basic version of the CheckboxInput is a unique `id`, a `label`, and an onchange event handler passed to the `onChange` prop.
```
juanfabrega, line 23: I believe the `isChecked` prop is also required here, right?

src/components/SelectInput/SelectInput.stories.mdx
```diff
@@ -0,0 +1,337 @@
...
+
+Allows users to pick a value from a predefined list of options. Ideally, it should be used when there are 5 or more options, otherwise you should consider using a radio group instead.
+
+## Props
+
+<Props of={SelectInput} />
+
+## Default
+
+All that is required to render a basic version of the SelectInput is a unique `id`, a `label`, a list of options passed to the `options` prop, and an onchange event handler passed to the `onChange` prop.
```
juanfabrega, line 23: Isn't the `value` prop required here as well?

PR #154, opened 2020-08-07
Consistent Bundling Refactor

JCarmine:
There was quite a bit of duplication between the production and storybook build files. It became apparent that this was dangerous when I attempted to add an auto-prefixer; I was essentially modifying the same block of code twice. But could have just as easily accidentally modified one, and not the other, causing the builds to get out of step.

So the solution, I believe, is a single webpack config file that modifies its config based on the env. I believe the output generated by this single file is identical to the two file solution, but please verify this.


webpack.config.js
```diff
@@ -0,0 +1,137 @@
...
+        loader: 'css-loader',
+        options: {
+          modules: true,
+          sourceMap: true,
+        },
+      },
+      'sass-loader',
+    ],
+    include: /\.module\.scss$/,
+  },
```
JCarmine, line 32: This is the block that was duplicated between the two files. Both versions had to be modified to install the autoprefixer.

PR #156, opened 2020-08-12
Storybook update

juanfabrega:
This PR updates to the latest version of Storybook, which makes our TS setup 100% no-config as far as storybook is concerned. 

I have also separated the creation of types (d.ts) into a `tsc` command so that the transpilation is fully done via `babel-loader` this allows us to generate proptypes which throw warnings for consumers who may still want runtime checks.


.storybook/main.js
```diff
@@ -22,7 +21,6 @@ module.exports = {
       ...config,
       plugins: [
         ...config.plugins,
-        ...webpackConfig.plugins,
```
JCarmine, line 25: If we no-longer need plugins from our webpack.config, then we can remove the logic that detects the ENV and conditionally generates the group of plugins. We can just straight define plugins as whatever is needed for the production build.

juanfabrega: @JCarmine good call. We still need to keep the css modules config rule but everything else is now outside of the conditional statement.

Issue #160, opened 2020-08-13
Allow SelectInput's selected value to be cleared


nickj-rp:
I need to be able to clear the selected value on a <SelectInput> component that does not have multi-select enabled.

We should just be able to pass the `isClearable` prop through to the underlying react-select component.

juanfabrega:
This works. I’ll draft a PR to expose more of our underlying react-select
props for use.

nickj-rp:
Many thanks!

juanfabrega:
![](https://github.trello.services/images/mini-trello-icon.png) [Clear SelectInput](https://trello.com/c/SDPp8D1v/166-clear-selectinput)

juanfabrega:
Open PR is here: https://github.com/palmetto/palmetto-components/pull/165

juanfabrega:
Hey @nickj-rp this has been merged and released in [release 0.2.0](https://github.com/palmetto/palmetto-components/releases/tag/v0.2.0)

Note I have already updated this dependency in the `logistics-platform` repo`master` branch as well.

PR #165, opened 2020-08-17
160 react select props

juanfabrega:
Added the `isClearable` prop to the SelectInput so users can clear the value as needed.


nathanyoung:
fixes #160

PR #173, opened 2020-08-19
Form label refactor

juanfabrega:
Refactored label component (and corresponding components that depend on it) to take children, rather than a `labelText' prop. The label prop on our inputs is still required to be a string for accessibility reasons.

### DIFF
--- a/src/components/CheckboxInput/CheckboxInput.tsx
+++ b/src/components/CheckboxInput/CheckboxInput.tsx
@@ -28,6 +28,15 @@ interface Props {
    */
   label: string;
   /**
+   * Determines if the label is not shown for stylistic reasons.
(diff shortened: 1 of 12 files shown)


src/components/CheckboxInput/CheckboxInput.tsx
```diff
@@ -82,34 +93,38 @@ const CheckboxInput: FC<Props> = ({
...
-              isFieldRequired: isRequired,
-              inputId: id,
-              labelText: label,
-              hasError: !!error,
-              isDisabled,
-            }}
-          />
+        <input {...inputProps} />
+        {label && !hideLabel && (
+          <FormLabel {...labelProps}>
```
RyanMacBern, line 125: props spreading seems unnecessary here and makes it harder to read in my opinion

PR #174, opened 2020-08-19
refactoring to use children components

juanfabrega:


src/components/Table/Table.stories.mdx
```diff
@@ -350,12 +357,12 @@ Setting the `isLoading` prop to true will disable clicking on the column headers
...
-        <Table
-          columns={columnConfig}
-          tableData={tableData}
-          useFixedWidthColumns
-          truncateOverflow
-        />
+        <Table useFixedWidthColumns>
+          <TableHead columns={columnConfig} />
+          <TableBody>
+            {tableData.map(row => <TableRow data={row} key={row.id} truncateOverflow useFixed/>)}
```
JCarmine, line 363: TableRow doesn't accept useFixed.

src/components/Table/Table.stories.mdx
```diff
@@ -4,6 +4,9 @@ import { withA11y } from '@storybook/addon-a11y';
 import { Meta, Story, Preview, Props } from '@storybook/addon-docs/blocks';
 import Button from '../Button/Button';
 import Table from './Table';
+import TableBody from './TableBody/TableBody';
+import TableHead from './TableHead/TableHead';
+import TableRow from './TableRow/TableRow';
```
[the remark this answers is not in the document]
JCarmine: For later, but since we are exposing these components now, should we expose them in Storybook?

src/components/Table/Table.stories.mdx
```diff
@@ -36,10 +39,12 @@ I'm a table!
...
       ];
       return (
-        <Table
-          columns={columnConfig}
-          tableData={tableData}
-        />
+        <Table>
+          <TableHead columns={columnConfig} />
+          <TableBody>
+            {tableData.map(row => <TableRow data={row} key={row.id} />)}
```
JCarmine, line 45: Console still has missing key errors. I think key needs to be received as a prop by TableRow, and this key applied to the `<td>` elements.

src/components/Table/Table.stories.mdx
```diff
@@ -428,11 +438,12 @@ Setting the `isLoading` prop to true will disable clicking on the column headers
         { id: 3, color: 'green', flavor: '' },
       ];
       return (
-        <Table
-          columns={columnConfig}
-          tableData={tableData}
-          emptyCellPlaceholder="--"
-        />
+        <Table isStriped>
```
JCarmine, line 441: Let's remove `isStriped` for this example to clarify that we are only demonstrating the placeholder.

src/components/Table/Table.tsx
```diff
@@ -52,39 +45,22 @@ interface Props {
...
   useFixedWidthColumns?: boolean;
 }
 
-const Table: FC<Props> = ({
-  columns,
-  tableData,
+const Table: FC<TableProps> = ({
   className = undefined,
-  emptyCellPlaceholder = undefined,
+  children = null,
```
JCarmine, line 57: Wouldn't children be required here? In this case, children referrers to TableBody, and only TableBody; there should never be a Table without a TableBody.

In this case, the <tbody> tag should probably be included in Table, to prevent someone from doing something foolish like forgetting it. 

Otherwise, we need a way to validate that the child is in fact the TableBody component.

src/components/Spinner/Spinner.tsx
```diff
@@ -3,7 +3,7 @@ import classNames from 'classnames';
 import { PALMETTO_BRAND_COLOR_VALUES, PALMETTO_COLORS } from '../../lib/tokens';
 import styles from './Spinner.module.scss';
 
-interface Props {
+interface SpinnerProps {
```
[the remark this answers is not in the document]
juanfabrega: Hey @JCarmine I am trying to boyscout these as I go. I know it is not ideal, but yes we want to move away from the non-descriptive props since this actually gets exported to a TS type and it is better to have the be specific.

src/components/Table/Table.stories.mdx
```diff
@@ -62,11 +67,12 @@ I'm a table!
...
       return (
         <>
-          <Table
-            columns={columnConfig}
-            tableData={tableData}
-            isLoading
-          />
+          <Table isLoading>
+            <TableHead columns={columnConfig} isLoading />
+            <TableBody isLoading>
```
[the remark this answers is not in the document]
juanfabrega: @JCarmine oversight. Updated.

src/components/Table/Table.stories.mdx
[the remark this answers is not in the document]
juanfabrega: Done, and yes! I think we should. I had planned to work on updated docs tomorrow.

src/components/Table/Table.stories.mdx
```diff
@@ -62,11 +67,12 @@ I'm a table!
       ];
       return (
         <>
-          <Table
-            columns={columnConfig}
-            tableData={tableData}
-            isLoading
-          />
+          <Table isLoading>
```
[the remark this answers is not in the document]
juanfabrega: I wondered this too, I guess at first glance I wasn't sure why the overlay wouldn't disable click events underneath it, feels like it should, but perhaps I'm missing something. Didn't want to fiddle with things too much so as not to break it  for this initial draft.

src/components/Table/Table.stories.mdx
```diff
@@ -129,12 +135,13 @@ I'm a table!
       }
       return (
         <Table
-          columns={columnConfig}
-          tableData={generateTableData()}
-          onSort={handleTableHeaderClick}
-          sortedColumn={sortedColumn}
           isResponsive
```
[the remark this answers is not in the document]
juanfabrega: Fair enough this one slipped through.

src/components/Table/Table.stories.mdx
juanfabrega: fixed!

src/components/Table/Table.stories.mdx
juanfabrega: Hmm the <tr> already have a key though. Checking.

src/components/Table/Table.tsx
juanfabrega: Not really since `TableHead` is outside of `TableBody` the table can take both as separate nodes. Not sure how to best describe this but it should be flexible that way.

PR #186, opened 2020-08-24
Modernize Legacy Components

JCarmine:
Most of this is standardizing imports, organizing the interfaces, cleaning up comments.

However, there are quite a few instances where I converted JSX to TSX. Could the TS folks really hammer on the TS bits?


src/components/Formik/FormikRadioGroup/FormikRadioGroup.tsx
```diff
@@ -0,0 +1,42 @@
+import React, {
+  FC,
+  ChangeEvent,
+  FocusEvent,
+} from 'react';
+import RadioGroup from '../../RadioGroup/RadioGroup';
+import { Option } from '../../RadioGroup/RadioGroupTypes';
```
JCarmine, line 7: These types ended getting shared between RadioGroup and the Formik component. Should there be a dir in the root to store these files? Or, should RadioGroup retain ownership of them?

Is the naming convention good?

RyanMacBern: I think it's reasonable for RadioGroup to retain ownership.

src/components/Formik/FormikSelectInput/FormikSelectInput.tsx
```diff
@@ -0,0 +1,48 @@
...
+    id: string;
+    label: string;
+    name: string;
+    options: Option[];
+    onBlur: (event: FocusEvent<HTMLElement>) => void;
+    onChange: (event: SimulatedEventPayload) => void;
+    //  eslint-disable-next-line @typescript-eslint/no-explicit-any
+    value?: { [key: string]: any; } | string | { [key: string]: any; }[];
+  };
+  form: { [key: string]: any; }; // eslint-disable-line @typescript-eslint/no-explicit-any
```
RyanMacBern, line 19: create a type for `{ [key: string]: any; }` rather than repeating it

src/components/RadioGroup/RadioInput/RadioInput.tsx
```diff
@@ -28,26 +29,26 @@ interface Props {
    * Mark the radio group as invalid and display a validation message.
    * Pass a string or node to render a validation message below the input.
    */
-  error?: boolean | string | React.ReactNode;
+  error?: boolean | string | ReactNode;
```
RyanMacBern, line 32: ReactNode includes boolean and string

src/components/RadioGroup/RadioGroupTypes.tsx
```diff
@@ -0,0 +1,6 @@
+export interface Option {
```
RyanMacBern, line 1: It would be reasonable to export this from RadioGroup.tsx instead of creating a RadioGroupTypes.tsx file.

src/components/Formik/FormikCheckboxInput/FormikCheckboxInput.tsx
```diff
@@ -0,0 +1,43 @@
...
+  ...props
+}) => (
+  <CheckboxInput
+    id={id}
+    label={label}
+    error={touched[name] && errors[name]}
+    isChecked={value}
+    onBlur={onBlur}
+    onChange={onChange}
+    {...props}
```
RyanMacBern, line 39: I'm not sure the two `...props` spreads are necessary at all.

src/components/Formik/FormikCheckboxInput/FormikCheckboxInput.tsx
```diff
@@ -0,0 +1,43 @@
...
+} from 'react';
+import CheckboxInput from '../../CheckboxInput/CheckboxInput';
+
+interface FormikCheckboxInputProps {
+  field: {
+    id: string;
+    label: string;
+    name: string;
+    onBlur: (event: FocusEvent<HTMLElement>) => void;
+    onChange: (event: ChangeEvent<HTMLInputElement>) => void;
```
ericyliu, line 14: should these be optional?

src/components/Formik/FormikCheckboxInput/FormikCheckboxInput.tsx
```diff
@@ -0,0 +1,43 @@
...
+    onChange,
+    value = false,
+  },
+  form: { touched, errors },
+  ...props
+}) => (
+  <CheckboxInput
+    id={id}
+    label={label}
+    error={touched[name] && errors[name]}
```
ericyliu, line 35: is it weird we're passing in the whole form object instead of an error directly?

src/components/Formik/FormikRadioGroup/FormikRadioGroup.tsx
ericyliu: i think because FormikRadioGroup is using RadioGroup, its ok for RadioGroup to own those types

src/components/Formik/FormikCheckboxInput/FormikCheckboxInput.tsx
```diff
@@ -0,0 +1,43 @@
...
+interface FormikCheckboxInputProps {
+  field: {
+    id: string;
+    label: string;
+    name: string;
+    onBlur: (event: FocusEvent<HTMLElement>) => void;
+    onChange: (event: ChangeEvent<HTMLInputElement>) => void;
+    value?: boolean;
+  };
+  form: { [key: string]: any; }; // eslint-disable-line @typescript-eslint/no-explicit-any
```
juanfabrega, line 17: Hey @JCarmine I could be wrong but I actually think that all of these params have type signatures that we can extend here. https://github.com/formium/formik/blob/master/packages/formik/src/types.tsx

src/components/Formik/FormikCheckboxInput/FormikCheckboxInput.tsx
JCarmine: @juanfabrega ?

src/components/RadioGroup/RadioGroupTypes.tsx
JCarmine: After experimenting with this in TextInput, I found that it really cluttered up the head of the file. I'm opting to keep the separate type files, to maintain simplicity in the actual components, and to make it obvious that components have types which might be used in other cases.

If this is ever the case, the type definition will be moved to /lib/types.tsx.

src/components/Formik/FormikCheckboxInput/FormikCheckboxInput.tsx
JCarmine: @juanfabrega

src/components/Formik/FormikCheckboxInput/FormikCheckboxInput.tsx
JCarmine: @juanfabrega

juanfabrega: Hey @RyanMacBern the component would not work properly otherwise. It is meant to pass our own component's (`<CheckboxInput>`) props down through the chain. They are merged with the formik field's props. The first line takes all of our extra props and assign's them to the variable `props`, and the second line passes them down. Without each of those statements, the process breaks down. 

perhaps it would be clearer to rename these as `rest` or `restProps` to make it clear that these are props in addition to the default formik field props. I'm good either way, but maybe that helps clarify the intent.

src/components/Formik/FormikCheckboxInput/FormikCheckboxInput.tsx
juanfabrega: @ericyliu the prop structure here is based on what the formik `<Field>` component actually passes down, we are only adhering to that structure and then spreading any of our custom props.

@JCarmine I noticed we added some props to the `field` object from what we had before, did you look at any docs that perhaps said these were missing? Formik uses the `name` prop to id a field, but I don't believe it utilizes the `id` or `label` those actually get passed down to our component as part of the prop spread. If you want to outline them individually we can do that outside of the `field` object.

src/components/Formik/FormikCheckboxInput/FormikCheckboxInput.tsx
juanfabrega: @ericyliu the props are passed down via Formik's `Field` component and that component passes down the form. We are documenting their signature in the proptypes to help library users, although in theory, we have no control over those default props (`form` and `field`) being passed down.

src/components/SelectInput/SelectInput.tsx
```diff
@@ -1,20 +1,20 @@
-import React, { FC, FocusEvent } from 'react';
+import React, { FC, FocusEvent, ReactNode } from 'react';
 import classNames from 'classnames';
 import Select, { ValueType, OptionTypeBase, FocusEventHandler } from 'react-select';
 import FormLabel from '../FormLabel/FormLabel';
 import InputValidationMessage from '../InputValidationMessage/InputValidationMessage';
 import styles from './SelectInput.module.scss';
 
-interface SimulatedEventPayload {
+type SimulatedEventPayloadType = {
```
JCarmine, line 8: I believe that this is just a type, that happens to be used in an interface, as well as elsewhere.

src/components/Table/TableTypes.ts
```diff
@@ -1,11 +1,11 @@
-export interface Column {
+export type ColumnType = {
```
JCarmine, line 1: Once again, I believe these are types, not interfaces.

src/lib/types.ts
```diff
@@ -0,0 +1,2 @@
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type UnknownPropertiesObjType = { [key: string]: any; };
```
JCarmine, line 2: @juanfabrega Is this where we want to keep the truly global types?

src/components/Button/Button.test.jsx
```diff
@@ -28,65 +28,91 @@ describe('Button', () => {
...
+        test(`it has a ${size.class} class applied to it`, () => {
+          render(<Button size={size.class}>{`${size.name} Button`}</Button>);
+
+          const btn = screen.getByText(`${size.name} Button`).closest('button');
+
+          expect(btn.getAttribute('class')).toContain(size.class);
+        });
+      })
+    ));
+  });
```
JCarmine, line 49: This was recommended by Code Climate, through an inline PR comment. Hooray for robots!

src/lib/types.ts
juanfabrega: @JCarmine yeah this should work.

src/components/Button/Button.test.jsx
juanfabrega: Heck yes! @JCarmine this is similar to how we wrote the `Heading` tests. I love it.

PR #201, opened 2020-09-01
Table Styling

nathanyoung:
- updates the styling of tables
- updates palmetto-design-tokens package
- updates the Default story example to a more applicable scenario of displaying a list of customers.
- moves all Table css vars to tables.css variables file


src/components/Table/Table.stories.mdx
```diff
@@ -55,37 +55,56 @@ NOTE: all properties in a column object are optional, but we highly recommend in
...
+        { heading: 'phone', dataKey: 'phone', width: 125 },
+        { heading: 'plan', dataKey: 'plan', width: 125 },
+        { heading: 'status', dataKey: 'status', width: 125 },
       ];
       const tableData = [
-        { id: 1, color: 'red', flavor: 'vanilla' },
-        { id: 2, color: 'blue', flavor: 'chocolate' },
-        { id: 3, color: 'green', flavor: 'strawberry' },
+        { status: 'lead captured',
+          customerInfo: { name: 'John Smith', email: 'john@palmetto.com' },
```
nathanyoung, line 87: @juanfabrega @JCarmine in order to combine properties of a "customer" object like name and email address into a single cell, do we have to combine them into an object first (in this example i'm combining them into a `customerInfo` object). 

I want to able to render tables like this

Screen Shot 2020-09-01 at 4 42 17 PM

src/components/Table/Table.module.scss
```diff
@@ -1,5 +1,4 @@
 .container {
-  display: inline-block;
```
nathanyoung, line 2: tables should be block elements. the parent should be responsible for setting it's width.

src/components/Table/Table.stories.mdx
```diff
@@ -55,37 +55,56 @@ NOTE: all properties in a column object are optional, but we highly recommend in
...
+        plan: 'standard'
+        },
       ];
       return (
-        <Table 
-          rowKey="id" 
-          columns={columnConfig} 
-          rows={tableData} 
+        <Table
+        useFixedWidthColumns
```
JCarmine, line 104: Indentation is off here.

src/components/Table/Table.stories.mdx
juanfabrega: @nathanyoung you do not have to combine them in your original data structure. Instead you can use the `render` method in the column config, which has access to the whole row. The render method will take 3 arguments `(cell, row, rowIndex)`. Here is an example of a column config: 

```
renderCustomer = (cell, row) => (
  <div>
    <p>{row.name}</p>
    <p className="text-grey-light">{row.email}</p>
  </div>
);

columns: [
 { heading: 'CUSTOMER', render: renderCustomer },
  ...restOfColumns
];
```

You can use the column render method to:
* Combine fields (forgoing the `dataKey` altogether.
* Shape data. E.G: transform primitive number values into currency strings.
* Style data. E.G: Create a button or link based on the cell data or row, or index.
* Render something completely unrelated to the data at hand, if you need to.

src/components/Table/Table.module.scss
```diff
@@ -16,16 +15,17 @@
 
   .table {
     cursor: default;
+    width: 100%;
     border-collapse: collapse;
     white-space: normal;
-    color: var(--color-font-base) !important;
+    color: var(--table-color) !important;
```
juanfabrega, line 21: Are we able to lose this !important here? I realize it was there before your PR but would be nice to get rid of it if we can.

PR #202, opened 2020-09-02
component-based reset

juanfabrega:
Here's a draft of the component-based css reset proposal.

**CONTEXT**
Some things I tried before arriving at this approach.

* A global CSS reset -- Unfortunately this is way to overzealous at this point in time and even if we did implement it I'm not sure we could control for all the variables that are at play. E.G: Other resets coming in from the likes of bootstrap or any number of the custom stylesheets that exist in our apps.
* A wrapper component for our components that imports a css reset. I.E: `<CssReset><ActualComponent></CssReset`. While this solution seemed elegant at first hand, it was riddled width issues as it was somewhat nightmarish in terms of extensibility (which styles to reset where). To get it functioning correct it also required a wrapper div for all components which I though was unnecessary.
* Importing the base reset into each component scss module. This is a step closer to the final solution but unfortunately led to all kinds of specificity issues as well as bleed into other areas.

In the end I wanted the solution to fill these criteria:
1. Does not target html elements globally.
2. Reset is strictly limited to the scope of the component being used and has no chance of bleeding into adjacent markup.
3. Reset is specific enough to ward off global stylesheets elsewhere in an application, but not too specific as to become inflexible in unique cases where custom styling is needed.
4. It tailors resets to the component at hand. This was something I decided along the way, but it seems much better to have a reset applied to the elements of a component, as opposed to just plastering every component with an entire. I.E, there is no need for us to reset `td` elements inside a our `FormLabel` component. This results in leaner bundles and more flexible components that only get resets on the properties they need.
5. Maintain the ability to keep a global reset (much leaner) that will apply to all components. For things like `font-family` and a few other attributes.

The basic approach is every component has a `component-name-reset` class, and that classes is composed off the `base-css-reset` class. From there the reset for each component can be extended based on the need.

If this looks good to everyone, I can go ahead and apply it across the board.


src/components/FormLabel/FormLabel.module.scss
```diff
@@ -1,7 +1,9 @@
+.label-reset {
+  composes: base-css-reset from '../../styles/reset.scss';
+}
+
 .label {
   display: block;
```
RyanMacBern, line 6: Why does it need to be a separate class? Why not put the `composes` right inside the `label` class?

juanfabrega: I find it helpful to know which properties on my component are being added purely for reset purposes vs things that are actually component styles.

juanfabrega:
FYI -- I noticed this also introduces some specificity issues. Reviewing some modifications.

src/styles/reset.scss
```diff
@@ -1,14 +1,8 @@
...
-h6 {
+.base-css-reset {
+  box-sizing: border-box;
   margin: 0;
-  font-weight: 600;
+  padding: 0;
+  line-height: normal;
+  font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen, Ubuntu, Cantarell, 'Open Sans', 'Helvetica Neue', sans-serif;
+  font-weight: 400;
 }
```
nathanyoung, line 8: if most of the components have their own .module.scss file with a `.componentName` class, why not set specfic reset values there instead of applying the same across the board?

Issue #21, opened 2020-07-01
Alert


nathanyoung:
For displaying error messages, such as when a form is submitted but the api returns an error.

MVP Proposed Props

| prop  | type | default value |  description |
| ------------- | ------------- | ------------- | ------------- |
| children  | node |  | content of the alert |
| autoHideDuration | number |  | number of milliseconds to wait before automatically dismissing the alert |
| dismissible | bool | false | when true, display a close button so the alert can be dismissed |
| onDismiss | func |  | callback function called when an alert is dismissed |

Post MVP

| prop  | type | default value |  description |
| ------------- | ------------- | ------------- | ------------- |
| type | enum | info | success, warning, danger, info |

nathanyoung:
![](https://github.trello.services/images/mini-trello-icon.png) [Alert Component](https://trello.com/c/yUvMfVif/51-alert-component)

PR #213, opened 2020-09-04
Alert Component

juanfabrega:
PR addressing this component: https://github.com/palmetto/palmetto-components/issues/21

Apologies, I moved on this without looking at the MVP props that are in the issue, but I believe the functionality is all there, with a few more options. The one exception being the `autoClose` on a timer, but that should be feasible without too much effort.


nathanyoung:
I'd like to keep the font color standard across all alerts. I started a figma file way back that looks like this:

Screen Shot 2020-09-04 at 10 57 15 AM

juanfabrega:
> I'd like to keep the font color standard across all alerts. I started a figma file way back that looks like this:
> 
> Screen Shot 2020-09-04 at 10 57 15 AM

Cool, I'm on it!

juanfabrega:
@nathanyoung this is ready for another onceover.

src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,154 @@
...
+  );
+
+  return (
+    <div className={alertContainerClasses} role="alert">
+      {hasIcon && renderAlertIcon()}
+      <div className={styles['alert-message']}>
+        {render ? render() : (
+          <>
+            {title && <Heading as="h4" size="md" className={styles['alert-title']}>{title}</Heading>}
+            {message && <span>{message}</span>}
```
nathanyoung, line 145: why a `span` and not a `p` or `div`?

src/components/Alert/Alert.module.scss
```diff
@@ -0,0 +1,59 @@
...
+    background-color: var(--alert-background-color-danger);
+    color: var(--alert-font-color-danger);
+  }
+
+  &.compact {
+    padding: var(--alert-padding-compact);
+  }
+
+  .type-icon {
+    margin-right: var(--size-spacing-sm);
```
nathanyoung, line 34: should we define a new alert token for this instead of using the spacing token directly?

src/components/Alert/Alert.module.scss
```diff
@@ -0,0 +1,59 @@
...
+
+
+    &.clickable {
+      cursor: pointer;
+    }
+  }
+
+  .alert-message {
+    &:not(:last-child) {
+      margin-right: var(--size-spacing-sm);
```
nathanyoung, line 50: should we define a new alert token for this instead of using the spacing token directly?

src/components/Alert/Alert.module.scss
```diff
@@ -0,0 +1,59 @@
...
+    }
+  }
+}
+
+.alert-title {
+  &:not(:last-child) {
+    margin-bottom: var(--size-spacing-2xs);
```
nathanyoung, line 57: should we define a new alert token for this instead of using the spacing token directly?

nathanyoung (changes requested):
Screen Shot 2020-09-08 at 10 22 11 PM

Can we color the icons so that they correspond with the type?

font-color-secondary for info
font-color-danger for danger,
font-color-success for success, etc...

src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,154 @@
...
+          <span
+            role="button"
+            onClick={handleClose}
+            onKeyUp={handleCloseKeyPress}
+            tabIndex={0}
+          >
+            {closeText}
+          </span>
+        ) : (
+          <FontAwesomeIcon
```
nathanyoung, line 120: Should be wrapped in a button. It's not keyboard focusable.

src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,159 @@
...
+          /* @TODO -- replace with <Button /> component once variations exists for this use case */
+          <button
+            type="button"
+            onClick={handleClose}
+            onKeyUp={handleCloseKeyPress}
+          >
+            {closeText}
+          </button>
+        ) : (
+          <FontAwesomeIcon
```
nathanyoung, line 125: The <FontAwesomeIcon /> needs to be wrapped in `<button>` too and the onClick handler moved. 

`role="button"` doesn't allow it to be focusable.

nathanyoung (changes requested):

PR #229, opened 2020-09-11
fix: Layout for sortable column headers

JCarmine:


src/components/Table/Table.module.scss
```diff
@@ -15,6 +15,7 @@
   }
 
   .table {
+    margin-top: 1px;
```
nathanyoung, line 18: @JCarmine Why is this necessary?

JCarmine: To make the focus ring rendered by Chrome fully visible; the ring's width is 2px, and it was partially obscured. We can chat about it in the am, and decide whether we care or not.

PR #235, opened 2020-09-15
Link

juanfabrega:
Implementation of a text link component.

It can be made to render either a react-router `Link` or a standard anchor tag based on whether the user passes a `href` or a `to` attribute.


nathanyoung (changes requested):
Screen Shot 2020-09-15 at 2 49 58 PM

I see `className` as a prop in TextLinkBaseProps but it's not showing up as a prop in the prop table documentation

nathanyoung:
`href` is also not displayed in the prop table

juanfabrega:
@nathanyoung I fixed the props issues. Our typescript docgen loader does some not so clear things when extending interfaces. See here: https://github.com/strothj/react-docgen-typescript-loader/issues/47

But with the new structure all 4 of the props we care about are there and the `...rest` should also be type safe.

nathanyoung (approved):

PR #243, opened 2020-09-19
Card Component

nathanyoung:


juanfabrega (approved):

src/components/Box/Box.tsx
```diff
@@ -257,9 +264,11 @@ const Box: FC<BoxProps> = ({
   const decorateChildren = (child: ReactElement, i: number, childrenArr: ReactElement[]) => {
     if (i === childrenArr.length - 1) return child; // Child gap does not apply to last element in the list.
 
-    const childClasses = classNames(child.props.className, childGapClass);
-
-    return cloneElement(child, { className: childClasses });
+    if (child && child.props) {
```
nathanyoung, line 267: @juanfabrega take a look at this. if a <Card.Header> didn't use the `title` prop, and only included children, this would break without this change.

PR #250, opened 2020-09-25
fix: Bug Fixes

JCarmine:


src/components/index.ts
```diff
@@ -1,31 +1,31 @@
+import Alert from './Alert/Alert';
```
JCarmine, line 1: We kind of lost touch on this file; quite a few things weren't correct. Might need to update our process to make sure it stays current.

src/docs/Spacing.stories.mdx
```diff
@@ -1,6 +1,6 @@
 import { Meta } from '@storybook/addon-docs/blocks';
 import sizes from '@palmetto/palmetto-design-tokens/build/js/variables-size';
-import { PALMETTO_SPACING_OPTIONS, PALMETTO_SPACING_SIZE_OPTIONS } from '../lib/tokens';
+import { PALMETTO_SPACING_OPTIONS, PALMETTO_SPACING_VALUES } from '../lib/tokens';
```
JCarmine, line 3: This story wasn't rendering in storybook, so I updated the constants being used. Did I use the right ones?

src/components/Alert/Alert.module.scss
```diff
@@ -71,15 +71,19 @@
...
     &:not(:last-child) {
       margin-right: var(--alert-message-margin);
     }
-  }
-}
 
-.alert-title {
-  &:not(:last-child) {
-    margin-bottom: var(--alert-title-margin);
+    > p {
```
JCarmine, line 85: Since .alert-message can only have one `<p>` as a direct descendent, I removed the bottom margin from it. However, if this `<p>` has any `<p>` descendents, they should still have their bottom margins applied, to maintain regular spacing.

nathanyoung:
Thanks for catching this stuff

nathanyoung (approved):

PR #253, opened 2020-09-26
docs(TextLink): accepts html anchor attributes

### LINKED ISSUE #249: TextLink documentation should say that it extends HTML anchor element

nathanyoung:
fixes #249

### DIFF
--- a/src/components/TextLink/TextLink.stories.mdx
+++ b/src/components/TextLink/TextLink.stories.mdx
@@ -16,6 +16,8 @@ or anchor tag based links.
 
 ## Props
 
+In addition to the props listed below, the `TextLink` accepts all valid HTML anchor element attributes, such as `_target`.
+
 <ArgsTable of={TextLink} />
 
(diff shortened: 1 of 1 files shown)


juanfabrega (approved):

src/components/TextLink/TextLink.stories.mdx
```diff
@@ -16,6 +16,8 @@ or anchor tag based links.
 
 ## Props
 
+In addition to the props listed below, the `TextLink` accepts all valid HTML anchor element attributes, such as `_target`.
```
nathanyoung, line 19:
```suggestion
When the `href` is set, the `TextLink` accepts all valid [HTML anchor element attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a), such as `_target`.

When `to` is set, the `TextLinik` accepts all [React Router Link](https://reactrouter.com/web/api/Link) props.
```

PR #264, opened 2020-10-07
fix: remove class specificity on list reset styles

nathanyoung:


src/styles/reset.scss
```diff
@@ -21,7 +21,7 @@ body {
 }
 
 /* Remove default padding */
-ul[class], ol[class] { /* stylelint-disable-line */
```
nathanyoung, line 24: this is preventing any of the utility classes from working.

juanfabrega (approved):

Issue #271, opened 2020-10-12
Chromatic runs twice after a branch is merged into main


nathanyoung:
After a pull request is merged, chromatic runs two builds on the `main` branch. It should only run once.  

![Screen Shot 2020-10-12 at 12 37 26 PM](https://user-images.githubusercontent.com/1447339/95783960-c16a0080-0c87-11eb-9a2a-b339e09d1c58.png)

PR #276, opened 2020-10-14
removing pr jobs on pushes

juanfabrega:
This removes the PR jobs from being run after merging to master. Also adds build job to run on PRs so that we are sure the build is running before firing a release.


nathanyoung (approved):
Let's find out if this works...

nathanyoung:
fixes https://github.com/palmetto/palmetto-components/issues/271

PR #278, opened 2020-10-15
Border sides + responsive colors on Box.

juanfabrega:
This PR includes the following features for the Box component.

NOTE: this is a breaking change since the `border` prop has been renamed to `borderColor`.

1. Border widths on specific sides.
2. Responsive border sizes and colors.
3. Responsive font colors.
4. Responsive background colors.

What we cannot do at the moment:
1. Different border colors on different sides.

### DIFF
--- a/src/components/Box/Box.tsx
+++ b/src/components/Box/Box.tsx
@@ -60,9 +60,9 @@ export interface BoxProps {
    */
   borderColor?: BrandColor | ResponsiveProp<BrandColor>;
   /**
-   * Any valid [brand color token](/?path=/docs/design-tokens-colors--brand) for the border color
+   * Any valid BorderSize token or reponsive prop with BorderSize in each breakpoint.
    */
-  borderWidth?: BorderSize;
+  borderWidth?: BorderSize | ResponsiveProp<BorderSize>;
   /**
    * Additional class names to add
    */
--- a/src/components/Box/Box.tsx
+++ b/src/components/Box/Box.tsx
@@ -56,8 +56,9 @@ export interface BoxProps {
   background?: BrandColor;
   /**
    * Any valid [brand color token](/?path=/docs/design-tokens-colors--brand) for the border color
+   * Or a responsive prop with BrandColor for each breakpoint.
    */
-  borderColor?: BrandColor;
+  borderColor?: BrandColor | ResponsiveProp<BrandColor>;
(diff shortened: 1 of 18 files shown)


src/components/Box/Box.tsx
```diff
@@ -38,25 +39,29 @@ export interface BoxProps {
...
    * Any valid [brand color token](/?path=/docs/design-tokens-colors--brand), or a `url()` for an image
    */
   background?: BrandColor;
   /**
    * Any valid [brand color token](/?path=/docs/design-tokens-colors--brand) for the border color
    */
-  border?: BrandColor;
+  borderColor?: BrandColor;
+  /**
+   * Any valid [brand color token](/?path=/docs/design-tokens-colors--brand) for the border color
```
nathanyoung, line 62: Needs to be updated and mention top, right, bottom, left convention

nathanyoung (changes requested):
Like we talked about, change `none` value to `0` for border width

nathanyoung:
Do the border width utility classes need to be more specific and use `border-width` instead of `border`? Just thinking about if we want to do something later with `border-style`

juanfabrega:
> Do the border width utility classes need to be more specific and use `border-width` instead of `border`? Just thinking about if we want to do something later with `border-style`

Hey @nathanyoung  this is strictly a question of how we generate the classes, I think. If we did border-widths, they would end up cancelling each other out in some cases for example if my borderWidth value was `xs sm` it would generate these classes.

```
.border-width-v-xs {
    border-width: 1px 0;
}

.border-width-h-sm {
    border-width: 2px 0;
}
```

As you can see these borders end up cancelling each other out and the box does not look the way I want it to.

The blocker (from my POV) is that if we wanted to do it this way we would have to generate a unique class for each possible combination of outputs (including 0) accounting for 1, 2, 3, and 4 value notation in CSS. So instead of having classes that smartly merge with each other to create a set of borders, we would be brute forcing every possible combination. I don't think that's a great idea and is set to grow exponentially the moment we add any further tokens to the border-width type.

juanfabrega:
[[Box] allow user to specify border side(s) and border width](https://trello.com/c/TWhzV27v/254-box-allow-user-to-specify-border-sides-and-border-width)

nathanyoung (approved):

PR #281, opened 2020-10-16
Add Formik Input Mask Examples.

JCarmine:


juanfabrega (approved):
Thanks @JCarmine. Looks great! 

Do you mind also adding just some text to the story, providing some specific context on the fact that the custom `onChange` function is a requirement for masked inputs? Just want to make sure that's 100% clear.

juanfabrega (approved):

PR #288, opened 2020-10-20
File upload

juanfabrega:
File upload component.

Screen Shot 2020-10-20 at 11 28 14 AM

### DIFF
--- a/src/components/Button/Button.test.jsx
+++ b/src/components/Button/Button.test.jsx
@@ -33,6 +33,7 @@ describe('Button', () => {
       { name: 'Small', class: 'sm' },
       { name: 'Medium', class: '' },
       { name: 'Large', class: 'lg' },
+      { name: 'XSmall', class: 'xs' },
     ];
 
     sizes.map(size => (
--- a/src/components/Button/Button.test.jsx
+++ b/src/components/Button/Button.test.jsx
@@ -142,9 +142,9 @@ describe('Button', () => {
         expect(buttonElement).toBeInTheDocument();
       });
 
-      test('it renders an anchor tag if an href attribute is passed as a prop', () => {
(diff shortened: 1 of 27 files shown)


nathanyoung (changes requested):
Looks good, but should we add it to the Formik Form example?

What about `isRequired`

juanfabrega:
> Looks good, but should we add it to the Formik Form example?
> 
> What about `isRequired`

Yup, definitely a good add. Will update and re-request review in a bit.

nathanyoung:
This story is rendering a `button` when it supposed to render a `<a>`

nathanyoung:
@juanfabrega The `isRequired` story results in this when pressing the submit button.

Screen Shot 2020-10-21 at 2 01 06 PM

src/components/Button/Button.test.jsx
```diff
@@ -142,9 +143,9 @@ describe('Button', () => {
         expect(buttonElement).toBeInTheDocument();
       });
 
-      test('it renders an anchor tag if an href attribute is passed as a prop', () => {
+      test('it renders an anchor tag if as prop `a` is passed', () => {
```
nathanyoung, line 146: this test is passing, but Chromatic says differently: https://www.chromatic.com/snapshot?appId=5ed9214b642dc10022b50a2d&id=5f9080a3409e4d0022907fcb

nathanyoung (changes requested):

src/components/Button/Button.test.jsx
juanfabrega: @nathanyoung I neglected to update the chromatic story, but the test has the correct implementation. Story is updating now.

nathanyoung (approved):

Issue #293, opened 2020-10-22
Modal


nathanyoung:
Modals are containers appearing in front of the main content to provide critical information or an actionable piece of content. These are specifically great for presenting new flows while preserving the state of content.

- on mobile there should be options to slide up from bottom and take over screen, or to be bottom aligned and only take up as much room as necessary
- focus trap
- supports any type of content
- close action
- close via esc key

PR #294, opened 2020-10-22
Modal

nathanyoung:


src/components/Modal/Modal.stories.jsx
```diff
@@ -0,0 +1,108 @@
+import React from 'react';
+import Modal from './Modal';
+import Button from '../Button/Button';
+
+export default {
+  title: 'Components/Modal/For Chromatic',
+  component: Modal,
+  parameters: {
+    chromatic: { delay: 1000 },
```
nathanyoung, line 9: @juanfabrega this doesn't seem to be working. I tried following these instructions:

https://www.chromatic.com/docs/delay#delay-all-stories-of-a-component

nathanyoung:
https://github.com/palmetto/palmetto-components/issues/293

src/components/Modal/Modal.module.scss
```diff
@@ -0,0 +1,107 @@
...
+:global([data-reach-dialog-overlay]) {
+  display: flex;
+  position: fixed;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  align-items: flex-end;
+  z-index: var(--size-z-index-overlay);
+  background: hsla(0, 0%, 0%, 0.33);
```
juanfabrega, line 44: Might be worth getting some tokens that represent semi-transparent colors. Not for this PR.

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,74 @@
+import React, { FC, ReactNode, RefObject } from 'react';
```
juanfabrega, line 1: Feels like we should allow prop spreading down to the DialogOverlay to cover any options that are exposed by reach.

juanfabrega (approved):

PR #295, opened 2020-10-22
Time picker

juanfabrega:
TimePicker component. Based on the SelectInput, but generates times based on start/end and interval.
Screen Shot 2020-10-22 at 3 35 45 PM

### DIFF
--- a/src/components/TimePicker/TimePicker.tsx
+++ b/src/components/TimePicker/TimePicker.tsx
@@ -13,7 +13,7 @@ interface TimePickerProps extends BaseSelectInputProps {
    */
   endTime?: { hour: number; minute: number; };
   /**
-   * Interval of displayed times (in minutes). Defaults to 15 minutes.
+   * Interval of displayed times (in seconds). Defaults to 900 seconds (15 minutes).
    */
   interval?: number;
   /**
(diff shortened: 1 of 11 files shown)


src/components/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,84 @@
...
+   * This is a direct passthrough to the second argument of JS `toLocaleTimeString`.
+   * [More](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString)
+   */
+  dateDisplayOptions?: Intl.DateTimeFormatOptions;
+  /**
+   * End hour and minute
+   */
+  endTime?: { hour: number; minute: number; };
+  /**
+   * Interval of displayed times (in minutes). Defaults to 15 minutes.
```
nathanyoung, line 16: This wording is a little confusing because you set the `interval` in seconds.

nathanyoung (approved):
Screen Shot 2020-10-22 at 2 32 19 PM

1. Should the `options`, prop be excluded since the times are autogenerated?
2. Should there be a FormikAdapter?

nathanyoung:
Also, might be good to have a story with the select open showing so that chromatic can capture the calculated options.

src/components/TimePicker/TimePicker.tsx
juanfabrega: I vacilated between the two which made the docs out of sync without me realizing. Fixed now.

PR #299, opened 2020-10-26
fix(Modal): allow overflow to be set by consumer

nathanyoung:


src/components/Box/Box.tsx
```diff
@@ -133,9 +133,7 @@ export interface BoxProps {
    */
   maxWidth?: DimensionSize | ResponsiveProp<DimensionSize> | string;
   /**
-   * The overflow property is specified as one or two keywords.
-   * If two keywords are specified, the first applies to overflow-x and the second to overflow-y.
```
nathanyoung, line 137: we aren't actually doing this

juanfabrega (approved):

PR #300, opened 2020-10-26
Sticky Header and Columns

JCarmine:


src/components/Table/TableCell/TableCell.module.scss
```diff
@@ -17,6 +18,13 @@
     white-space: nowrap;
   }
 
+  &.sticky {
+    position: sticky;
+    left: 0;
+    font-size: var(--table-header-font-size);
+    font-weight: 600;
```
JCarmine, line 25: This was to match the font-weight in the header. While the weight is appropriate for the use case, we don't have it in the design tokens. Should it be added? 
Screen Shot 2020-10-26 at 3 03 03 PM

src/components/Table/TableHeaderCell/TableHeaderCell.module.scss
```diff
@@ -27,6 +28,20 @@
     white-space: nowrap;
   }
 
+  &.sticky-header {
+    position: sticky;
+    top: 0;
+
+    &.sticky-column {
+      z-index: var(--size-z-index-sticky);
```
JCarmine, line 36: Believe this is what this was intended for?

nathanyoung: yes!

src/components/Table/TableCell/TableCell.tsx
```diff
@@ -57,19 +62,33 @@ const TableCell: FC<TableCellProps> = ({
...
-    <td
-      className={tableCellClasses}
-      style={{ ...width && { minWidth: `${width}px`, maxWidth: `${width}px` } }}
-    >
-      {(children === null || typeof children === 'undefined' || children === '') ? emptyCellPlaceholder : children}
-    </td>
+    <>
+      {isColumnSticky
+        ? (
+          <th
```
nathanyoung, line 76: since a `th` can be used for a column or row, we should add the `scope` property. 

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/th

nathanyoung (commented):
I pushed up some styling tweaks, but i'm still thinking about how we can visually show that a table has been scrolled horizontally.

i have one comment on regarding adding a `scope` property to the `th`.

Looks great so far!

src/components/Table/TableCell/TableCell.tsx
JCarmine: Good catch! Will update this evening and add some tests tomorrow a.m.

JCarmine:
> I pushed up some styling tweaks, but i'm still thinking about how we can visually show that a table has been scrolled horizontally.
> 
> i have one comment on regarding adding a `scope` property to the `th`.
> 
> Looks great so far!

Regarding styling, it seems like we should visually indicate that a column or row is sticky (special). In spreadsheets (Sheets, Excel, etc), this is typically done by increasing the border width between the sticky and non-sticky content.

If it were only possible to sticky the first column, I would just set `border-right` to `2px` or `3px` on the sticky column and call it done. However, in the case where, say, the 5th column is sticky, we should have a way to indicate this to the consumer before they ever scroll.

juanfabrega (changes requested):
@JCarmine excited for this and the PR is looking great, nice work!

My only questions RE: functionality are:

1. can we make the fixed column be an option of `right` or `left`? I think it would be helpful to have the ability to make columns on the right side sticky as well, in cases where perhaps we have an _actions_ column or something similar on the right side of the table. 
2. Currently it doesn't seem possible to sticky more than one column at the same time, in terms of use cases, I think we should discuss with @nathanyoung but it might be helpful in some cases to sticky more than one column.

Open to thoughts on this.

src/components/Table/Table.stories.mdx
```diff
@@ -390,6 +390,173 @@ with the second column having a fixed width of 300 pixels.
...
+        { id: 6, color: 'green', flavor: 'strawberry', brightness: 'dark' },
+        { id: 7, color: 'green', flavor: 'strawberry', brightness: 'dark' },
+      ];
+      return (
+        <div style={containerStyle}>
+          <Table
+            rowKey="id"
+            columns={columnConfig}
+            rows={tableData}
+            scr
```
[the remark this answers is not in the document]
nathanyoung: > > I pushed up some styling tweaks, but i'm still thinking about how we can visually show that a table has been scrolled horizontally.
> > i have one comment on regarding adding a `scope` property to the `th`.
> > Looks great so far!
> 
> Regarding styling, it seems like we should visually indicate that a column or row is sticky (special). In spreadsheets (Sheets, Excel, etc), this is typically done by increasing the border width between the sticky and non-sticky content.
> 
> If it were only possible to sticky the first column, I would just set `border-right` to `2px` or `3px` on the sticky column and call it done. However, in the case where, say, the 5th column is sticky, we should have a way to indicate this to the consumer before they ever scroll.

Right, I'll try playing with adding box-shadow to see if that works for the left most and the nth column.

nathanyoung:
> @JCarmine excited for this and the PR is looking great, nice work!
> 
> My only questions RE: functionality are:
> 
> 1. can we make the fixed column be an option of `right` or `left`? I think it would be helpful to have the ability to make columns on the right side sticky as well, in cases where perhaps we have an _actions_ column or something similar on the right side of the table.

Agree with ☝🏽 

> 2. Currently it doesn't seem possible to sticky more than one column at the same time, in terms of use cases, I think we should discuss with @nathanyoung but it might be helpful in some cases to sticky more than one column.

I think only having 1 sticky column is fine. The workaround for two sticky columns is combining the data into a single cell.

> 
> Open to thoughts on this.

src/components/Table/TableBody/TableBodyCell/TableBodyCell.test.tsx
```diff
@@ -14,7 +14,7 @@ describe('TableCell', () => {
       expect(tableCell).toBeInTheDocument();
     });
 
-    test('td element is rendered with specific width style atrtibute based on width prop', () => {
```
JCarmine, line 17: This extension is priceless: https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker

src/components/Table/TableBody/TableBodyCell/TableBodyCell.tsx
```diff
@@ -24,6 +24,14 @@ interface TableCellProps {
...
    */
   isCompact?: boolean;
+  /**
+   * Will stick to left side of table during horizontal scroll.
+   */
+  isColumnStickyLeft?: boolean;
+  /**
+   * Will stick to right side of table during horizontal scroll.
+   */
+  isColumnStickyRight?: boolean;
```
nathanyoung, line 34: instead of having two separate booleans, recommend changing this to a single prop that accepts `left | right`. This prevents accidentally having both `isColumnStickyLeft` and `isColumnStickyRight` at the same time.

nathanyoung (changes requested):
combine the column isSticky... props into a single prop

nathanyoung:
This story that adds custom classes no longer works.

https://5ed9214b642dc10022b50a2d-ywrcbaekrk.chromatic.com/?path=/story/components-table--custom-column-classes

https://www.chromatic.com/snapshot?appId=5ed9214b642dc10022b50a2d&id=5f98e87577f5750022ba0063

nathanyoung (dismissed):
Looks great! Don't forget to add the `sticky` prop to the Column prop table.

Screen Shot 2020-10-28 at 12 32 51 PM

src/components/Table/TableHead/TableHeaderCell/TableHeaderCell.module.scss
```diff
@@ -27,6 +28,30 @@
...
 
+  &.sticky-header {
+    position: sticky;
+    top: 0;
+    z-index: var(--size-z-index-sticky);
+
+    &.sticky-column-left,
+    &.sticky-column-right {
+      z-index: calc(var(--size-z-index-sticky) + 1);
+      box-shadow: 0 9px 0 0 var(--table-background-color), 0 -9px 0 0 var(--table-background-color), 2px 0 5px 0 rgba(0, 0, 0, 0.05), 1px 0 1px 0 rgba(0, 0, 0, 0.05), -2px 0 5px 0 rgba(0, 0, 0, 0.05), -1px 0 1px 0 rgba(0, 0, 0, 0.05);
```
juanfabrega, line 39: Is this not throwing a stylelint error?

juanfabrega (approved):

nathanyoung (changes requested):
see https://www.chromatic.com/build?appId=5ed9214b642dc10022b50a2d&number=1111

nathanyoung (approved):

PR #305, opened 2020-10-28
test: adds breakpoints snapshots to capture responsive Box stories in Chromatic

nathanyoung:
For each "responsive" story, this pr adds a chromatic parameter to capture snapshots at different viewport widths. Because of the way storybook puts stories into an iframe, i had to add 20px to each of our breakpoint values. They end up being: 

- 320 - iphone 5
- 700 (680) - tablet
- 1012 (992) - desktop
- 1300 (1280) - hd

### DIFF
--- a/src/components/Box/Box.stories.mdx
+++ b/src/components/Box/Box.stories.mdx
@@ -806,7 +806,7 @@ Set the maximum width of the Box by setting `maxHeight` to a [height token](/?pa
 
 ### Overflow
 
-Set the overlfow behavior of the Box with `overflow`
+Set the overflow behavior of the Box with `overflow`
 
 <Canvas>
   <Story name="overflow">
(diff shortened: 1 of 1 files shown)


src/components/Box/Box.stories.mdx
```diff
@@ -864,7 +867,10 @@ The box will have different margin and padding values for at the base, tablet, a
 #### Direction
 
 <Canvas>
-  <Story name="responsive direction">
+  <Story
+    name="responsive direction"
+    parameters={{ chromatic: { viewports: [320, 700, 1012, 1300] } }}
```
nathanyoung, line 872: @juanfabrega not sure if there's a way to use the `BREAKPOINT_VALUES` defined in tokens.js and add 20, so the viewports array is just hardcoded.

juanfabrega: @nathanyoung I do believe there is a better way to do this, and make it available globally as a storybook decorator/setting. It would take maybe 1-2 hours between implementation and testing, I think. For now if you want to merge as is I don't think that's an issue.

nathanyoung: going to do the above as a separate change.

juanfabrega (approved):

PR #306, opened 2020-10-28
feat(Popover): MVP version of Popover component.

juanfabrega:
Ready for review.


src/components/Popover/Popover.module.scss
```diff
@@ -0,0 +1,76 @@
+/* stylelint-disable */
+.popover {
+  box-shadow:
```
nathanyoung, line 3: There's a token for this!

src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,204 @@
...
+const Popover: FC<PopoverProps> = ({
+  arrowColor = undefined,
+  children,
+  content,
+  contentContainerProps = { ...contentContainerDefaults },
+  isOpen,
+  onClickOutside = undefined,
+  placement,
+  portalTarget = document.body,
+  trapFocus = false,
```
nathanyoung, line 83: Should this be true by default?

src/components/Popover/Popover.module.scss
juanfabrega: will address before merging.

src/components/Popover/Popover.tsx
juanfabrega: Hey @nathanyoung having the `trapFocus = true` by default is not a problem but it does create a big pain for anybody using our components and running tests. The tests fail to run properly and require an `await` placed strategically, which even after a dev gets it right (ends up delaying tests substantially).

nathanyoung: ok

nathanyoung (approved):

PR #315, opened 2020-10-30
fix: Bring Back the Stripes!

JCarmine:


src/components/Table/TableBody/TableBody.module.scss
```diff
@@ -1,13 +1,24 @@
 .table-body {
+  th,
+  td {
+    background-color: var(--table-background-color);
+  }
```
JCarmine, line 5: Doesn't technically need to be here, but, I wanted all of the background colors for tbody declared in one place. Since the striped and hoverable rules HAVE to be here, it all goes here.

juanfabrega (approved):

PR #320, opened 2020-11-03
fix: if a user specifies '0' as a spacing value, return the corresponding css className

nathanyoung:
We have utility classes that set margin and padding values to 0, but we don't use them currently because the `cssShorthandToClasses` function only returns classnames if a value is not equal to zero or a border width. 

This prevents us from setting a spacing value to 0 at different breakpoints (see linked issue)

This change always returns a css classname if the spacing value is valid.

### DIFF
--- a/src/lib/cssShorthandToClasses.ts
+++ b/src/lib/cssShorthandToClasses.ts
@@ -33,10 +33,7 @@ function generateBaseClasses(
     const sides = value.split(' ');
 
     sides.forEach((v, index) => {
-      // Exception for border width since we want to default borders to zero.
-      if (v !== '0' || attribute === 'border-width') {
-        classes.push(`${attribute}-${shorthand[sides.length][index]}-${v}`);
-      }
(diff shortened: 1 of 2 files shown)


src/lib/cssShorthandToClasses.ts
```diff
@@ -33,10 +33,7 @@ function generateBaseClasses(
     const sides = value.split(' ');
 
     sides.forEach((v, index) => {
-      // Exception for border width since we want to default borders to zero.
-      if (v !== '0' || attribute === 'border-width') {
-        classes.push(`${attribute}-${shorthand[sides.length][index]}-${v}`);
-      }
+      classes.push(`${attribute}-${shorthand[sides.length][index]}-${v}`);
```
juanfabrega, line 36: Since this exception is specifically for border-width, I'm unclear how this affects spacing classes. is there something I'm missing?

juanfabrega: nevermind, I misread this and now it makes perfect sense. @nathanyoung

juanfabrega (approved):

PR #331, opened 2020-11-17
Nexa font

juanfabrega:
Adding nexa font assets as well as making it the default for Headings.


src/components/Heading/Heading.module.scss
```diff
@@ -1,3 +1,4 @@
 .heading {
   line-height: var(--size-line-height-heading);
+  font-family: Nexa, -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen, Ubuntu, Cantarell, 'Open Sans', 'Helvetica Neue', sans-serif;
```
nathanyoung, line 3: should this come from a token instead?

juanfabrega: tokens for fonts are not available at the moment.

nathanyoung (approved):

PR #352, opened 2020-12-05
SSR Build

juanfabrega:
This PR switches our build process so that it does not inline module CSS and instead writes it to an `index.css` file. 

NOTE that this will require users to add this new index.css file to their projects as noted in the documentation.


nathanyoung:
Let's double check and make sure this works with the logistics app if you haven't already

Issue #356, opened 2020-12-08
Modal footer unreachable in mobile (Android)


JCarmine:
This is a result of setting 100vh on the modal's container. While the height of the modal IS 100vh, unfortunately, the calculation does not take into consideration the height of the address bar on mobile. This leads to a case where the fixed element cannot be scrolled too.

Due to this bug, we should probably never use 100vh on things, which is unfortunate, since it would be so handy if it worked correctly.

![2020_12_08_13 11 33](https://user-images.githubusercontent.com/5695990/101543470-245ed780-3959-11eb-8ed1-6bc7f8208d9b.jpg)

juanfabrega:
Thanks for bringing this up @JCarmine 

Some potential solutions:

1. Use window.innerHeight, which should have the actually height of the viewport.
2. Have the modal footer be absolutely or sticky positioned with bottom: 0; While keeping the modal content scroll:auto by default.

For solution 1, we'd need to make sure we check for DOM availability so we don't break SSR.

Perhaps there is a third option I'm not thinking of.

juanfabrega:
cc: @nathanyoung

JCarmine:
I have encountered this issue in the past. In that particular case, there was nothing fixed to the bottom, but, you still had to scroll unnecessarily to reach the end of the page; more annoying than anything.

Turns out the solution was as simple as just setting height 100% on the html, body, etc.

Link to the app where I first discovered the 100vh issue (view on Android to note that the issue does not occur, despite the presence of the address bar): https://www.jessecarmine.com/demos/react-weather-app/

Repo: https://github.com/JCarmine/react-weather-app
https://www.jessecarmine.com/demos/react-weather-app/

PR #358, opened 2020-12-08
fix(Modal): fix for footer modal on android/ios browsers

nathanyoung:


nathanyoung:
https://github.com/palmetto/palmetto-components/issues/356

PR #366, opened 2020-12-17
fix: dont reset ol, ul list types because browser default is ok

nathanyoung:
Because of our reset.scss, using an order list `ol` element renders a bullet list instead of numbered list. 

This change fixes that by removing the `list-style-type` property, and relying on the browser default to set the list type.

### DIFF
--- a/src/styles/reset.scss
+++ b/src/styles/reset.scss
@@ -19,15 +19,16 @@ body {
   /* Core body defaults */
   text-rendering: optimizeSpeed;
   /* Setting font family and default px value (to be used by our rem units downstream) */
-  font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen, Ubuntu, Cantarell, 'Open Sans', 'Helvetica Neue', sans-serif;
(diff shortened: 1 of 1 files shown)


src/styles/reset.scss
```diff
@@ -19,15 +19,16 @@ body {
   /* Core body defaults */
   text-rendering: optimizeSpeed;
   /* Setting font family and default px value (to be used by our rem units downstream) */
-  font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen, Ubuntu, Cantarell, 'Open Sans', 'Helvetica Neue', sans-serif;
+  font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen, Ubuntu, Cantarell,
```
nathanyoung, line 22: white space

src/styles/reset.scss
```diff
@@ -68,7 +69,8 @@ figure,
 figcaption,
 blockquote,
 dl,
-dd { /* stylelint-enable */
+dd {
+  /* stylelint-enable */
```
nathanyoung, line 73: white space only

juanfabrega (approved):

PR #392, opened 2021-01-28
Box stories

nathanyoung:
- Reorganizes and consolidates Box component stories to new format.
- fixes background and fontColor prop type definitions
- adds box playground story

Screen Shot 2021-01-29 at 5 50 12 PM


src/types/index.ts
```diff
@@ -122,6 +122,7 @@ export type CssTextAlignValue =
   'right';
 
 export type BaseSpacing = SpacingSize | string | undefined;
+export type BorderRadiusSize = BorderRadius | string | undefined;
```
nathanyoung, line 125: like spacing types, this accepts css shorthand (e.g. `xs 0 sm lg`)

PR #398, opened 2021-02-12
docs: Heading component documentation re-organization

osbornedesign:
image

### DIFF
--- a/src/components/Heading/Heading.Overview.stories.mdx
+++ b/src/components/Heading/Heading.Overview.stories.mdx
@@ -72,7 +72,7 @@ The `size` prop accepts an object where the properties correspond to [breakpoint
 The variant of the heading is inherited from its parent, but can be set to brand variant as well.
 
 <Canvas>
-  <Story name="Colors">
+  <Story name="Variant">
     <>
       {FONT_COLOR_OPTIONS.map((variant) => (
         <Heading as="h3" variant={variant} key={uuid()}>
--- a/src/components/Heading/Heading.Overview.stories.mdx
+++ b/src/components/Heading/Heading.Overview.stories.mdx
@@ -1,9 +1,9 @@
-import Heading from "./Heading";
-import { v4 as uuid } from "uuid";
-import { action } from "@storybook/addon-actions";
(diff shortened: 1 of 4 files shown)


osbornedesign:
@nathanyoung can you check the Visual Regression Tests and make sure this is how we want to display them. I wasn't sure how to go about breaking them into sections but then was questioning if that is even necessary

nathanyoung:
> @nathanyoung can you check the Visual Regression Tests and make sure this is how we want to display them. I wasn't sure how to go about breaking them into sections but then was questioning if that is even necessary

It looks like you've got 1 too many loops and the visual regression test story. There are so many loops that chromatic isn't even rendering it because it's too long.

nathanyoung:
Screen Shot 2021-02-12 at 12 50 37 PM

I think we want a story for
- each size
- each color
- each size and color combination
- responsive size

osbornedesign:
> > @nathanyoung can you check the Visual Regression Tests and make sure this is how we want to display them. I wasn't sure how to go about breaking them into sections but then was questioning if that is even necessary
> 
> It looks like you've got 1 too many loops and the visual regression test story. There are so many loops that chromatic isn't even rendering it because it's too long.

It looks like there are too many loops but it's because of the "level" prop. It is showing each size as "h1, h2, h3..."
image

src/components/Heading/Heading.Overview.stories.mdx
```diff
@@ -71,7 +74,7 @@ The variant of the heading is inherited from its parent, but can be set to brand
 <Canvas>
   <Story name="Colors">
```
nathanyoung, line 75:
```suggestion
  <Story name="Variant">
```

src/components/Heading/Heading.VisualTests.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+    ))}
+  </Box>
+);
+
+export const Size = SizeTemplate.bind({});
+Size.args = {
+  children: 'Hello world!',
+};
+
+export const Color = ColorTemplate.bind({});
```
nathanyoung, line 45:
```suggestion
export const Variant = ColorTemplate.bind({});
```

nathanyoung (approved):

PR #399, opened 2021-02-13
fix(Button): don't pass type property to button anchor links

nathanyoung:


src/components/Button/Button.module.scss
```diff
@@ -4,6 +4,7 @@
   line-height: 1;
 
   &:not(.naked) {
+    display: inline-block;
```
nathanyoung, line 7: applying this because when rendering an anchor tag, we want it to have inline-block behavior.

PR #407, opened 2021-02-23
feat: add underline, strikethrough, utility classes

nathanyoung:


nathanyoung:
closing until we wire up props for this, otherwise they are undocumented utility classes.

Issue #357, opened 2020-12-08
Cursor displays in DateInput, but nothing can be entered


JCarmine:
It seems correct that nothing can be entered, forcing the user to interact with the calendar. However, there are two UX issues resulting from this:

1) Even though nothing can be directly entered, the cursor still flashes, leading the user to think that it could.
2) On mobile, focusing on DateInput will prompt the mobile keyboard to open, which obscures the calendar. Since nothing can be directly entered, the keyboard popping just makes the feature more difficult to use.

**A possible solution:** Adding the `readonly` attribute to the input.

![2020_12_08_13 15 41](https://user-images.githubusercontent.com/5695990/101545481-2a09ec80-395c-11eb-92f8-4d67e78a7789.jpg)

juanfabrega:
@JCarmine just an FYI that we have pulled this issue into our current sprint on the UX side.

nathanyoung:
related: https://github.com/palmetto/palmetto-components/issues/376

juanfabrega:
https://github.com/palmetto/palmetto-components/pull/419

PR #419, opened 2021-03-04
fix(DateInput): Make date input read only.

juanfabrega:
This PR addresses this issue:
https://github.com/palmetto/palmetto-components/issues/357

* we have made the dateinput readonly so that it does not trigger the mobile keyboard.
* we've added another default mask to allow for entering dates easily into a text input.

### DIFF
--- a/src/components/TextInput/TextInput.Overview.stories.mdx
+++ b/src/components/TextInput/TextInput.Overview.stories.mdx
@@ -381,6 +381,29 @@ Use the `type` prop to mark the input type as "tel". Use the `inputMask` prop to
   </Story>
 </Canvas>
 
+## With Date Mask
+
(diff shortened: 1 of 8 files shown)


src/components/TextInput/TextInput.Overview.stories.mdx
```diff
@@ -381,6 +381,29 @@ Use the `type` prop to mark the input type as "tel". Use the `inputMask` prop to
...
+
+<Canvas>
+  <Story name="With Date Mask">
+    {() => {
+      const [value, setValue] = useState('');
+      return (
+        <TextInput
+          id="withDateMask"
+          value={value}
+          label="Label"
```
nathanyoung, line 396:
```suggestion
          label="Date of Birth"
```

nathanyoung (approved):
One suggestion, otherwise looks good

PR #439, opened 2021-03-31
Box additional states

juanfabrega:
# Github Issue or Trello Card
This PR addresses this issue: https://trello.com/c/WCfLgkvQ/538-box-hover-states

This change adds hover and focus states for the box component. It ALSO adds `cursor` utility classes which go hand-in-hand with hover.

- [X] New feature (non-breaking change which adds functionality)
- [ ] Updating documentation.

# Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes

# UI Checklist
- [X] I have conducted visual UAT on my changes/features.
- [X] My solution works well on desktop, tablet, and mobile browsers.


src/components/Box/Box.tsx
```diff
@@ -244,7 +285,34 @@ export const Box: FC<BoxProps> = forwardRef((
...
+    : undefined;
+
+  const focusClasses = focus
+    ? Object.entries(focus).map(([key, value]) => (
+      cssPropertyMap[key].transformer(`focus:${cssPropertyMap[key as keyof BoxProps['focus']].classPrefix}`, value)
+    ))
+    : undefined;
+
   const boxClasses = classNames(
+    styles.box,
```
nathanyoung, line 315: I think we should only apply the transition styles when a hover or focus is set.

src/components/Box/Box.module.scss
```diff
@@ -0,0 +1,16 @@
+.box {
```
nathanyoung, line 1: instead of making this a box css module, should it be a transition utility?

nathanyoung (changes requested):
The hover border example doesn't seem to be working right, it only shows a bottom border.

https://user-images.githubusercontent.com/1447339/113184231-cad13000-9209-11eb-889e-64ca21dbdbd2.mov

src/components/Box/Box.module.scss
juanfabrega: yes, I think that would be ideal, I just felt that the scope of the pr was getting a bit away from me with adding focus and cursor utilities on top of hover.

I wasn't 100% sure how to best move forward with transition utilities given that they should probably allow transition of specific properties and not necessarily across the board when applied.

src/components/Box/Box.tsx
juanfabrega: fari enough.

juanfabrega:
> The hover border example doesn't seem to be working right, it only shows a bottom border.
> 
>  hover-border.mov

@nathanyoung this was done on purpose, kinda like an underline, but if it is confusing I can change it.

PR #461, opened 2021-05-05
feat: allow popover to be server-side rendered

mtblanton:
# Github Issue or Trello Card
This PR addresses this issue: https://trello.com/c/bC292LqA

Palmetto.com has to force some clients to only render on the client-side instead of being server-side rendered. This fixes that for any components using Popover directly, but not through DateInput (DateInput forces withPortal to true, which renders it with a portalTarget of document.body).

LMK what you think of the typing, esp around the discriminated union. It prohibits setting portalTarget unless withPortal is true. I did this because we only ever need portalTarget if withPortal is true. I wanted to make it only required if withPortal is true but couldn't do that with the param initialization!

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Checklist:

          The typing for Popover isn't quite the same. Let me know if that's a problem.

# UI Checklist
        the storybook still looked good/worked! P.com proposal also still looks good.

### DIFF
--- a/src/components/Popover/Popover.tsx
+++ b/src/components/Popover/Popover.tsx
@@ -18,7 +18,7 @@ import { BrandColor } from '../../types';
 import styles from './Popover.module.scss';
 import { Box, BoxProps } from '../Box/Box';
 
-export interface PopoverProps {
+export type PopoverProps = {
   /**
    * The trigger element
    */
@@ -62,15 +62,16 @@ export interface PopoverProps {
    */
   placement?: Placement;
   /**
-   * The target element where the Popover will be portaled to, when `withPortal === true`.
-   */
-  portalTarget?: HTMLElement;
-  /**
    * Whether you want to trap focus in the Popover element when it is open.
    * Read more about focus traps:
    * [Here](https://allyjs.io/tutorials/accessible-dialog.html#trapping-focus-inside-the-dialog)
    */
   trapFocus?: boolean;
+  /**
+   * Additional props to be spread to rendered element
+   */
+  [x: string]: any; // eslint-disable-line
+} & ({
   /**
    * Whether the element should be rendered outside its DOM structure
    * for reasons of placement. Use this when the element is being cut-off or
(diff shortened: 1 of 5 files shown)


src/components/Popover/Popover.tsx
```diff
@@ -61,29 +61,33 @@ export interface PopoverProps {
...
-  /**
-   * Additional props to be spread to rendered element
+  withPortal: true;
+   /**
+   * The target element where the Popover will be portaled to, when `withPortal === true`.
    */
-  [x: string]: any; // eslint-disable-line
-}
+  portalTarget?: HTMLElement;
+} | {
```
mtblanton, line 87: See https://basarat.gitbook.io/typescript/type-system/discriminated-unions for more info on how this works!

src/components/Popover/Popover.tsx
```diff
@@ -101,9 +105,9 @@ export const Popover: FC<PopoverProps> = ({
   offsetFromTarget = 12,
   onClickOutside = undefined,
   placement = 'right',
-  portalTarget = document.body,
-  trapFocus = false,
   withPortal = false,
+  portalTarget = withPortal ? document.body : undefined,
```
mtblanton, line 109: This has potential to still cause issues in SSR contexts if withPortal is enabled, but I think Portals cause problems anyway so that's a problem for another day!

src/components/Popover/Popover.tsx
```diff
@@ -61,29 +61,33 @@ export interface PopoverProps {
...
-  withPortal?: boolean;
-  /**
-   * Additional props to be spread to rendered element
+  withPortal: true;
+   /**
+   * The target element where the Popover will be portaled to, when `withPortal === true`.
    */
-  [x: string]: any; // eslint-disable-line
-}
+  portalTarget?: HTMLElement;
```
juanfabrega, line 86: In this case, do we still need to mark this as optional with the `?` feels like it should be required if `withPortal` is true.

juanfabrega (approved):
Looks, great thanks! Just had a small question regarding the type interface but other than that this is good to go. Appreciate the contribution as always.

As you noted the evaluation of document.body likely still creates some issues. We could also NOT set that default at all, since that prop is required when `withPortal` is true, in which case not specifying falls under user error from my POV.

src/components/Popover/Popover.tsx
mtblanton: Having it default to document.body seemed nice! Having that default was the only thing keeping me from making portalTarget required, so if you want to get rid of that default we can make this required!

mtblanton: Juan and I discussed this over Slack, we're moving forward with the change to require portalTarget if withPortal is true

mtblanton:
[proposal controls not showing up](https://trello.com/c/bC292LqA/1068-proposal-controls-not-showing-up)

PR #469, opened 2021-05-12
feat: add support for cross platform development

osbornm:
- [x] Updating deployment/build pipeline.

## The issue
The current package scripts assume a unix (MacOS, linux, etc.) shell, via the way they set environment variables and call `rm`.

## Why does it matter
I am aware that everyone, on the main team, is likely using a unix style shell but in the event that someone else wants to help out or fix a bug and they are running, for example windows, they don't have to figure out these issues and they can focus on their contribution not the infrastructure. Additionally, the overhead for using these packages is in the order of 10s of milliseconds so it should not effect anyone already using a unix style shell.  

## The fix
Use two new dev packages `rimraf` and `cross-env` to perform environment variable sets and deletes of folders. 

# Checklist:

- [x] (N/A) I have commented my code, particularly in hard-to-understand areas
- [x] (N/A) I have made corresponding changes to the documentation
- [x] (N/A) I have added tests that prove my fix is effective or that my feature works

# UI Checklist 
- [x] (N/A) I have conducted visual UAT on my changes/features.
- [x] (N/A) My solution works well on desktop, tablet, and mobile browsers.


juanfabrega (approved):

juanfabrega:
appreciate the contribution @osbornm

PR #497, opened 2021-06-08
Updating value type to match available option types

CharlesPlucker:
# Overview

When selecting a utility provider and tariff, we receive the tarriffId as a number. When I pass in that number into the value field of the SelectInputNative, I get a type error that it doesn't match a string:

image

I believe SelectInputNative should accept a value that matches the types of the option's keys.

# Github Issue or Trello Card
This PR addresses this issue: 
https://trello.com/c/74yuGx9I/1943-utility-rate-selection-limited-to-the-first-entry

- [x] Updating documentation.

Before Doc: 
image
After Doc: 
image

# Checklist:

# UI Checklist

### DIFF
--- a/src/components/SelectInputNative/SelectInputNative.tsx
+++ b/src/components/SelectInputNative/SelectInputNative.tsx
@@ -8,7 +8,7 @@ export interface SelectInputNativeBaseProps extends BoxProps, FormControlProps {
   /**
    * Value of selected option. Should match the value key in the option object.
    */
-  value: string | null;
+  value: string | number | null;
   /**
    * onChange callback from select element.
    */


osbornm:
**NOTE** tests are a little weird with this type of change since they are "design/develop" time type errors. I couldn't think of a good way to test this.

juanfabrega:
Hey @CharlesPlucker at first glance this seems like an oversight (on our part) in the option interface as opposed to the value prop type. 

Per the HTML spec an option's value should be a DOMString: https://html.spec.whatwg.org/multipage/form-elements.html#the-option-element 

My first instinct is to honor that API, but I'm open to making this flexible provided it doesn't introduce any major errors. Let me know what you think.

juanfabrega:
> **NOTE** tests are a little weird with this type of change since they are "design/develop" time type errors. I couldn't think of a good way to test this, but not too worried about it.

osbornm:
@juanfabrega by the time the value gets to the HTML is will be a string, react handles converting to for us. you'd just want to stop things like object from being converted to something like `[object]` as the value :)  

### React
```
{() => {
      const [value, setValue] = useState(1);
      const options = [
        { value: 1, label: 'Chocolate' },
        { value: 'strawberry', label: 'Strawberry' },
        { value: 'vanilla', label: 'Vanilla' },
      ];
      return (
        <div>
          <SelectInputNative
            id="default Select"
            label="Label"
            onChange={event => { setValue(event.target.value) }}
            options={options}
            value={value}
          />
        </div>
      );
    }}
```
### HTML
```
<select class="flex-direction-column display-flex font-size-inherit font-color-dark" aria-label="Label" aria-labelledby="default SelectLabel" id="default Select">
  <option class="flex-direction-column display-flex font-size-inherit" value="" disabled="" hidden="">Select...</option>
  <option class="flex-direction-column display-flex font-size-inherit" value="1">Chocolate</option>
  <option class="flex-direction-column display-flex font-size-inherit" value="strawberry">Strawberry</option>
  <option class="flex-direction-column display-flex font-size-inherit" value="vanilla">Vanilla</option>
</select>
```

juanfabrega (approved):

CharlesPlucker:
[Utility Rate selection limited to the first entry](https://trello.com/c/74yuGx9I/1943-utility-rate-selection-limited-to-the-first-entry)

CharlesPlucker:
[Utility Rate selection limited to the first entry](https://trello.com/c/74yuGx9I/1943-utility-rate-selection-limited-to-the-first-entry)

PR #530, opened 2021-07-15
Wrap required input label's asterisk in a targetable element

jedfoster:
# Github Issue or Trello Card
This PR addresses this issue: https://github.com/palmetto/helios-app/pull/231

On the app's login page, we wanted to require the Email and Password fields, but _not_ show the asterisk in the labels, as it's generally understood that all fields in a login form are required.

Wrapping the asterisk in a required field's label with a `span` and a class makes it a cinch to hide it with CSS, without requiring additional prop plumbing.

# Checklist:

# UI Checklist

### DIFF
--- a/src/components/FormLabel/FormLabel.test.jsx
+++ b/src/components/FormLabel/FormLabel.test.jsx
@@ -2,6 +2,16 @@ import React from 'react';
 import { render, screen } from '@testing-library/react';
 import { FormLabel } from './FormLabel';
 
+function getByTextWithMarkup(text) {
+  return (content, element) => {
+    const hasText = node => node.textContent === text;
+    const elementHasText = hasText(element);
+    const childrenDontHaveText = Array.from(element.children).every(child => !hasText(child));
+
+    return elementHasText && childrenDontHaveText;
+  };
+}
+
 beforeEach(() => {
   console.error = jest.fn(); // eslint-disable-line no-console
 });
@@ -24,7 +34,9 @@ describe('FormLabel', () => {
         my label
       </FormLabel>,
     );
-    const labelElement = screen.getByText('my label *');
+
+    const labelElement = screen.getByText(getByTextWithMarkup('my label *'));
+
     expect(labelElement).toBeInTheDocument();
   });
 
--- a/src/components/FormLabel/FormLabel.tsx
+++ b/src/components/FormLabel/FormLabel.tsx
@@ -75,7 +75,7 @@ export const FormLabel: FC<FormLabelProps> = ({
       {...restProps}
     >
       {children}
-      {isFieldRequired && <>&nbsp;*</>}
+      {isFieldRequired && <span className="field-required-asterisk">&nbsp;*</span>}
       {helpText && (
         <Box as="p" display="block" fontSize="sm" color="grey" className={styles['help-text']}>
           {helpText}
--- a/src/components/FormLabel/FormLabel.tsx
+++ b/src/components/FormLabel/FormLabel.tsx
@@ -75,7 +75,7 @@ export const FormLabel: FC<FormLabelProps> = ({
       {...restProps}
     >
       {children}
-      {isFieldRequired && <span className="field-required-asterisk">&nbsp;*</span>}
+      {isFieldRequired && <span> *</span>}
       {helpText && (
         <Box as="p" display="block" fontSize="sm" color="grey" className={styles['help-text']}>
           {helpText}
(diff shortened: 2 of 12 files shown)


jedfoster:
Hmmm. I see there was originally a span here, https://github.com/palmetto/palmetto-components/pull/286/files#diff-6a73775d173eddab177529270ff40bdba01f9881058ca9d6288571dd403a5fcaL67, why was this removed?

nathanyoung:
@JCarmine can you shed in light on why we removed the <span> from way back in October 2020? https://github.com/palmetto/palmetto-components/pull/286/files#diff-6a73775d173eddab177529270ff40bdba01f9881058ca9d6288571dd403a5fcaL67

juanfabrega (changes requested):
Hey @jedfoster as always, thanks for the contribution. We'd like to provide this flexibility given your use case, but I feel an empty class purely for targeting the asterisk seems less than ideal.

Would you be open to making this a prop (something like `hideAsterisk`) in the component to show or hide it as needed?

EDIT: another alternative is accepting more values for the `isRequired` prop so that you can specify if the asterisk is shown.

E.G

`isRequired: boolean | 'no-asterisk'` where no asterisk makes a field required (keeping its aria attributes) but does not show the asterisk

nathanyoung:
Alternatively, what if we allowed the `aria-required` value to be set by the user? That's the only other thing besides rendering the asterisk that `isRequired` is responsible for

jedfoster:
I'd be fine with either no class on the span or a prop. I originally figured a span (class or no) would be the easiest as it doesn't require any additional prop plumbing, but I seem to have broken a bunch of tests and fixing them is proving harder than expected.

jedfoster:
> Alternatively, what if we allowed the aria-required value to be set by the user? That's the only other thing besides rendering the asterisk that isRequired is responsible for

On a related note, I was going to open an issue or a PR about the fact that `isRequired` doesn't set the `required` attribute on the input, only `aria-required`.

jedfoster:
An classless span would be enough for my CSS-targeting purposes.

juanfabrega:
@jedfoster if a span works lets do with that for now. Let me know if you need help getting the tests sorted

jedfoster:
@juanfabrega Here's a failing test with that asterisk in a span:

```
 FAIL  src/components/FormLabel/FormLabel.test.jsx
  FormLabel
    ✓ Label correctly renders with base props (28ms)
    ✕ Label correctly renders with askterisk if field is required (9ms)
    ✓ correctly assigns an id when given an inputId (3ms)
    ✓ renders help text if provided (5ms)

  ● FormLabel › Label correctly renders with askterisk if field is required

    TestingLibraryElementError: Unable to find an element with the text: my label *. This could be because the text is broken up by multiple elements. In this case, you can provide a function for your text matcher to make your matcher more flexible.

    <body>
      <div>
        <label
          class="label m-0 p-0 display-block font-size-inherit"
          for="myId"
          id="myIdLabel"
        >
          my label
          <span>
             *
          </span>
        </label>
      </div>
    </body>

      25 |       </FormLabel>,
      26 |     );
    > 27 |     const labelElement = screen.getByText('my label *');
         |                                 ^
      28 |     expect(labelElement).toBeInTheDocument();
      29 |   });
      30 |

      at Object.getElementError (node_modules/@testing-library/dom/dist/config.js:37:19)
      at node_modules/@testing-library/dom/dist/query-helpers.js:90:38
      at node_modules/@testing-library/dom/dist/query-helpers.js:62:17
      at getByText (node_modules/@testing-library/dom/dist/query-helpers.js:111:19)
      at Object.<anonymous> (src/components/FormLabel/FormLabel.test.jsx:27:33)
```

I can modify the test like so and get it to pass:

```diff
-    const labelElement = screen.getByText('my label *');
+    const labelElement = screen.getByText((content, element) => {
+      const hasText = node => node.textContent === 'my label *';
+      const elementHasText = hasText(element);
+      const childrenDontHaveText = Array.from(element.children).every(child => !hasText(child));
+
+      return elementHasText && childrenDontHaveText;
+    });
```

There are a dozen or so other tests that will need updating as well.

But there's a dragon hiding in there... Because of the `&nbsp;` in `<span>&nbsp;*</span>`,  `'my label *' !== 'my label *'`. The space in front of the asterisk is actually `%C2%A0` instead of `%20`. A subtle difference that I don't want to put into the tests.

The alternative is `<span> *</span>`, which visually seems fine in Storybook.

Thoughts?

juanfabrega:
@jedfoster go for it!

nathanyoung:
@jedfoster Just want to let you know we didn't forget about this. This change impacts the Toggle component that I'm trying to work through. It doesn't have the correct space and alignment in certain scenarios. Chromatic actually picked up the changes. 

Screen Shot 2021-07-22 at 2 03 13 PM
Screen Shot 2021-07-22 at 2 03 06 PM

nathanyoung (approved):

PR #538, opened 2021-07-23
fix(TextLink): allow font size utility classes to be applied

nathanyoung:
# Github Issue or Trello Card

Allows css utility font size classes to get applied rather than always inheriting from the parent, for example:

`<TextLink className="font-size-sm font-size-md-tablet" />`

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Updating documentation.

# Checklist:

# UI Checklist
- [ ] I have conducted visual UAT on my changes/features.
- [ ] My solution works well on desktop, tablet, and mobile browsers.

### DIFF
--- a/src/components/TextLink/TextLink.module.scss
+++ b/src/components/TextLink/TextLink.module.scss
@@ -1,6 +1,4 @@
 .text-link {
-  font-size: inherit;
-
   &.primary {
     text-decoration-color: var(--text-link-font-color-primary);
     color: var(--text-link-font-color-primary);


src/components/TextLink/TextLink.module.scss
```diff
@@ -1,6 +1,4 @@
 .text-link {
-  font-size: inherit;
```
nathanyoung, line 2: Inheriting font size is already the default behavior, so the style is redundant, but also prevents usage of css utility classes

Issue #732, opened 2022-03-15
ButtonProps allows adding invalid props to DOM Elements


dg-harris:
The `[key: string]: any` on ButtonProps prevents typescript from properly typechecking Props on the Button Element.

PR #733, opened 2022-03-15
Fix: [732] Implement Stricter ButtonProps

dg-harris:
# Github Issue or Trello Card
This PR addresses this issue:  #732

This PR gives the Button component a more strict type signature to prevent accidental adding of react props to DOM elements. While this wouldn't adversely affect the dotcom team, I'm not familiar with how other teams are using this library so let me know if this would be too much trouble for other users.

I've kept the old `ButtonProps` type signature for backward compatibility in the case where others were using this type for their own components. If it would be preferred to make this a breaking change just to only have one set of props let me know.

# Completeness Checklist

- [x] TESTS: My changes maintain the baseline required test coverage, as specified by code climate analysis.
- [x] My code has no linting or typescript compile warnings.
- [x] My work is tied to a Github issue and satisfies the acceptance criteria (if applicable) of the corresponding issue.

# Quality Checklist:

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -1,5 +1,5 @@
 import React, {
-  FC, ReactNode, MouseEvent, FocusEvent, forwardRef, createElement, AnchorHTMLAttributes,
+  ReactNode, MouseEvent, FocusEvent, forwardRef, createElement, AnchorHTMLAttributes,
 } from 'react';
 import classNames from 'classnames';
 import { IconName, ResponsiveProp } from '../../types';
@@ -99,6 +99,9 @@ interface BaseButtonProps {
       * The color variant of the button
       */
      variant?: ButtonVariant;
+     /**
+      * ref - currently cannot be typed due to limitations of using the `as` prop
+      */
 }
 
 /**
@@ -122,7 +125,7 @@ type NormalButtonProps = { as?: 'button'; } &
 
 export type StrictButtonProps = NormalButtonProps | AnchorButtonProps
 
-export const Button: FC<StrictButtonProps> = forwardRef(
+export const Button = forwardRef<HTMLAnchorElement | HTMLButtonElement, StrictButtonProps>(
   (
     {
       children = undefined,
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -14,101 +14,115 @@ import styles from './Button.module.scss';
 export type ButtonVariant = 'primary' | 'success' | 'danger' | 'light' | 'dark' | 'white';
 
 export type ButtonSize = 'xs' | 'sm' | 'md' | 'lg';
-export interface ButtonProps {
-  /**
-   * Contents of the button.
-   */
-  children?: ReactNode;
+
+interface BaseButtonProps {
+     /**
+      * Contents of the button.
+      */
+     children?: ReactNode;
+     /**
+      * Additional ClassNames to add to button.
+      */
+     className?: string;
+     /**
+      * Button takes up the full width of its parent container.
+      */
+     fullWidth?: boolean;
+     /**
+      * Name of the icon to include before the button text
+      */
+     iconPrefix?: IconName;
+     /**
+      * Name of the icon to include after the button text
+      */
+     iconSuffix?: IconName;
+     /**
+      * A unique identifier for the button.
+      */
+     id?: string;
+     /**
+      * URL to navigate to when clicked. Passing this attribute automatically
+      * renders an anchor <a> tag, NOT a <button> element.
+      */
+     href?: string;
+     /**
+      * Disables the button, making it inoperable.
+      */
+     isDisabled?: boolean;
+     /**
+      * Replaces the button text with a loading indicator and disables the button.
+      */
+     isLoading?: boolean;
+     /**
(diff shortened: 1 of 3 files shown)


juanfabrega:
Hey @dg-harris I agree with implementing stricter props on everything moving forward. The original intent of the [x: any] was to allow things like data attributes, but I think these may be covered by react's base html prop interfaces.

I think my only question on this one is, why create a `StrictButtonProps` and not just make the stricter version the default ButtonProps?

dg-harris:
> 

For the new `StrictButtonProps` type my goal was to avoid breaking anyone that was using the existing `ButtonProps` within their own codebase. I think that's unlikely but wanted at least my first proposal to be as safe as possible

I could absolutely see just rolling this into a larger breaking change that makes all the props more strict for a major version increase. I'm good to close this and I'd be happy to pitch on a single larger change for more strict props.

RyanMacBern:
This is great. I agree the `[x: any]` should be removed from all the components. I don't expect that to cause any issues for the Palmetto App.

If desired, we can use the HTML element props instead. Although, as Taylor pointed out, I'm not sure how well that will work with the `as` prop.

mtblanton:
I have no clue how emotion's `styled` does this, but this kind of setup would be perfect:

image

relevant bit: `DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>`

juanfabrega:
@dg-harris perhaps other folks disagree, but I think trying to tackle this across the library might be more problematic in terms of getting this out in the wild and tested. I'm generally in favor of doing these refactors bit by bit, but I can understand wanting to bundle it in a breaking change. I think whether this is a breaking change is still to be seen though, since it appears that we don't have any (or frequent) uses of custom props in most components.

Curious to get others' opinions on this.

mtblanton:
I think this would break .com, but it'd be a good break. We have some prop spreads and `styled` calls that leak props through to P-C components. We'd love to get those leaks called out with TS so they're easier to recognize and fix!

dg-harris:
> > The original intent of the [x: any] was to allow things like data attributes, but I think these may be covered by react's base html prop interfaces.
> 
> I want to ask about what @juanfabrega brought up about data attributes. In [GLIDE](https://github.com/palmetto/glide), we're using `data-testid` attributes on Buttons. Will this still be supported?

Sinc

> I have no clue how emotion's `styled` does this, but this kind of setup would be perfect:
> 
> image
> 
> relevant bit: `DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>`
> 
> Example using something stringy like "as": image
> 
> all this said, I don't this this is a stopper. I'd like to see something like this in the future, though. Otherwise any non-anchor/button use will need a @ts-ignore



> I have no clue how emotion's `styled` does this, but this kind of setup would be perfect:
> 
> image
> 
> relevant bit: `DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>`
> 
> Example using something stringy like "as": image
> 
> all this said, I don't this this is a stopper. I'd like to see something like this in the future, though. Otherwise any non-anchor/button use will need a @ts-ignore

the `as` prop was already typed as `'button' | 'a'` so I think this should be good. Something like this wouldn't work for components like Box though.

juanfabrega:
@dg-harris what is your level of comfort just switching to the strict props as a default for this component to get this one out there as a starting point? I.E: getting rid of `StrictButtonProps` and just making the `ButtonProps` strict and extend the relevant HTML element attributes?

dg-harris:
It looks like the only other component that would really need this is Box.

When I looked into doing this for box using this kind of union method across

> 

Sounds good to me! I'll go ahead and make that change as well.

I actually found out that refs weren't working so I added one more change to fix refs if you want to take another look. They aren't perfect (you could pass an HTMLButton ref to a button rendered as an anchor and vis-versa) but the typing is more accurate than before this change before.

RyanMacBern (approved):

src/components/Button/Button.tsx
```diff
@@ -14,101 +14,105 @@ import styles from './Button.module.scss';
...
+      */
+     variant?: ButtonVariant;
+     /**
+      * ref - currently cannot be typed due to limitations of using the `as` prop
+      */
 }
 
-export const Button: FC<ButtonProps> = forwardRef(
+type AnchorButtonProps = { as: 'a'; } &
+  BaseButtonProps & Omit<JSX.IntrinsicElements['a'], 'ref'>
```
juanfabrega, line 108: Hey @dg-harris I saw your comment and understand these were perhaps problematic before, but I am not 100% following how this works. Won't omitting the ref attribute from JSX props type mean that passing a `ref` to the Button will now throw a TS error?

I'm sure I'm missing something but having some difficulty understanding this one.

dg-harris: Ya no problem, it threw me off to. The issue is that the intrinsicElement refs are all `LegacyRef` but that's not compatible with `forwardRef`

The key here is that I've removed the explicit typing on `Button`  and added types to `forwardRef` so the `ref` prop actually comes from the `forwardRef` type signature.

The `FC` type actually doesn't allow refs so this was already a problem, it was just being masked by `[x: string]: any`

juanfabrega (approved):

PR #737, opened 2022-03-22
feat(Card): enable one-off styling through BoxProps while maintaining theming

nathanyoung:
# Github Issue or Trello Card
This PR addresses this issue: https://trello.com/c/AeUi6Yvb/1209-card-if-a-box-prop-is-used-do-not-apply-the-card-classes

## Current State

With the recent implementation of component design tokens to enable theming, currently there is no way to allow for single-instance (or one-off) styling of a Card and it's child components. This is due to the fact that the classes applied by the Card.module.scss will always have a higher level of specificity than the utility classes that get applied via BoxProps.

## Solution

This pr makes it so that using any BoxProp whose css attribute matches the Card's will take precendent over the component design token's css attribute. This allows users of the component to use the default styles, use customize the appearance of their cards using the component design tokens, or have one-off styled versions of the Card using BoxProps.

Here's a screenshot of the documentation that illustrates this:

Screen Shot 2022-03-22 at 3 41 23 PM

```jsx
// default appearance
<Card>
  <Card.Section>Default appearance</Card.Section>
  <Card.Section>Default appearance</Card.Section>
  <Card.Footer>footer</Card.Footer>
</Card>

// customized card theme
<Box
  childGap="xl"
  style={{
    '--card-background-color': 'var(--color-brand-tertiary-50)',
    '--card-footer-background-color': 'var(--color-brand-tertiary-900)',
    '--card-border-radius': 'var(--size-border-radius-lg)',
    '--card-section-border-color': 'var(--color-brand-tertiary-100)',
    '--card-box-shadow': '0',
  }}
>
// themed card instance
  <Card>
    <Card.Section>Themed card using component design tokens</Card.Section>
    <Card.Section>Themed card using component design tokens</Card.Section>
    <Card.Footer color="white">footer</Card.Footer>
  </Card>
// one-off styled card within a themed container
  <Card shadow="xl" radius="sm" background="secondary-50">
    <Card.Section>BoxProps are used instead of tokens</Card.Section>
    <Card.Section borderWidth="sm 0 0 0" borderColor="secondary-100">
      BoxProps are used instead of tokens
    </Card.Section>
    <Card.Footer background="secondary-900" color="grey-100">
      footer
    </Card.Footer>
  </Card>
</Box>
```

- [x] Updating documentation.
- [ ] Updating Project Dependencies
- [ ] Improving or adding to Test Coverage

# Completeness Checklist

- [ ] TESTS: My changes maintain the baseline required test coverage, as specified by code climate analysis.
- [ ] DOCS: All new component work is covered in that component's `Overview` docs.
- [ ] DOCS: All new component work is covered in a component `Playground`.
- [ ] DOCS: All new visual changes or options are covered under relevant components' `VisualTests`.
- [ ] My code has no linting or typescript compile warnings.
- [ ] My work is tied to a Github issue and satisfies the acceptance criteria (if applicable) of the corresponding issue.

# Quality Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] My changes generate no new warnings
- [ ] New and existing unit tests pass locally with my changes

# UI Checklist
- [ ] I have conducted visual UAT on my changes/features.
- [ ] My solution works well on desktop, tablet, and mobile browsers.

### DIFF
--- a/src/components/Card/Card.Overview.stories.mdx
+++ b/src/components/Card/Card.Overview.stories.mdx
@@ -317,6 +317,29 @@ Here are some common use cases of `<Card />`.
   </Story>
 </Canvas>
 
+## Single-Instance Styling
+
+Occasionally, you may want to customize a single instance of a `Card`, but still have the ability to manage the appearance of all other Card instances through its component design token values. We allow single-instance overrides through the use of BoxProps.
+
+**Using any BoxProp whose css attribute matches a component design token's attribute, will mean that all the component's design token values will be ignored. It is assumed that a greater granularity of visual control is desired.**
+
+<Canvas>
+  <Story name="Single-Instance Styling">
+    <Box childGap="xl">
+      <Card>
+        <Card.Section>Default appearance</Card.Section>
+        <Card.Footer>footer</Card.Footer>
+      </Card>
+      <Card shadow="xl" radius="md" background="secondary-50">
+        <Card.Section>Customized appearance using BoxProps</Card.Section>
+        <Card.Footer background="secondary-900" color="grey-100">
+          footer
+        </Card.Footer>
+      </Card>
+    </Box>
+  </Story>
+</Canvas>
+
 ## Component Design Tokens
 
 <table>
--- a/src/components/Card/Card.Overview.stories.mdx
+++ b/src/components/Card/Card.Overview.stories.mdx
@@ -332,6 +332,7 @@ Here are some common use cases of `<Card />`.
         '--card-background-color': '--color-brand-white-500',
         '--card-border-radius': '--size-border-radius-md',
         '--card-box-shadow': '--size-box-shadow-sm',
+        '--card-subdued-background-color': '--color-brand-grey-50',
         '--card-section-border-color': '--color-brand-grey-100',
         '--card-footer-background-color': '--color-brand-grey-50',
       };
--- a/src/components/Card/Card.Overview.stories.mdx
+++ b/src/components/Card/Card.Overview.stories.mdx
@@ -321,17 +321,19 @@ Here are some common use cases of `<Card />`.
 
 Occasionally, you may want to customize a single instance of a `Card`, but still have the ability to manage the appearance of all other Card instances through its component design token values. We allow single-instance overrides through the use of BoxProps.
 
-**Using any BoxProp whose css attribute matches a component design token's attribute, will mean that all the component's design token values will be ignored. It is assumed that a greater granularity of visual control is desired.**
-
 <Canvas>
   <Story name="Single-Instance Styling">
     <Box childGap="xl">
       <Card>
         <Card.Section>Default appearance</Card.Section>
+        <Card.Section>Default appearance</Card.Section>
         <Card.Footer>footer</Card.Footer>
       </Card>
       <Card shadow="xl" radius="sm" background="secondary-50">
         <Card.Section>Customized appearance using BoxProps</Card.Section>
+        <Card.Section borderWidth="sm 0 0 0" borderColor="secondary-100">
+          Customized appearance using BoxProps
+        </Card.Section>
         <Card.Footer background="secondary-900" color="grey-100">
           footer
         </Card.Footer>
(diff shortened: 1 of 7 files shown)


src/components/Card/Card.module.scss
```diff
@@ -1,6 +1,6 @@
 .card {
   background-color: var(--card-background-color, var(--color-brand-white-500));
-  border-radius: var(--modal-border-radius, var(--size-border-radius-md));
+  border-radius: var(--card-border-radius, var(--size-border-radius-md));
```
nathanyoung, line 3: bug!

src/components/Card/Card.Overview.stories.mdx
```diff
@@ -332,6 +368,7 @@ Here are some common use cases of `<Card />`.
         '--card-background-color': '--color-brand-white-500',
         '--card-border-radius': '--size-border-radius-md',
         '--card-box-shadow': '--size-box-shadow-sm',
+        '--card-subdued-background-color': '--color-brand-grey-50',
```
nathanyoung, line 371: adding to documentation. was left out before

src/components/Card/Card.tsx
```diff
@@ -13,38 +21,61 @@ export interface CardProps extends BoxProps {
...
+      overflow = 'hidden',
+      display = 'block',
+      radius = undefined,
+      shadow = undefined,
+      width = '100',
+      ...restProps
+    },
+    ref,
+  ) => {
+    const useTheme = background === undefined && radius === undefined && shadow === undefined;
```
nathanyoung, line 52: Background, radius, and shadow are themeable component design tokens, so if any of these are passed as props, don't apply the theme styles.

arempe93:
does passing a `background` prop also remove theme styling for `border-radius`?

I think we can split the classes up to keep those defaults

nathanyoung:
@arempe93 yes, at the moment it would strip out all theming styling. I’ll separate it like you suggest, that’s a better way to handle it

nathanyoung:
@arempe93 @juanfabrega I think this is ready for a another review. I added a code example (it's the same as the in the new documentation) in the PR description

src/components/Card/Card.module.scss
```diff
@@ -1,19 +1,30 @@
...
   }
 }
 
-.card-footer {
+.card-footer-background {
   background-color: var(--card-footer-background-color, var(--color-brand-grey-50));
-  border-color: var(--card-section-border-color, var(--color-brand-grey-100));
 }
 
-.card-section + .card-section {
```
juanfabrega, line 17: Why is this not necessary moving forward?

src/components/Card/components/CardFooter/CardFooter.tsx
```diff
@@ -1,34 +1,60 @@
...
 export interface CardFooterProps extends BoxProps {
   /**
    * Contents of the Footer.
    */
   children?: ReactNode;
+  /**
+   * If defined as a prop, this value will take higher precedence than the corresponding component design token value
+   * Any valid [brand color token](/?path=/story/design-tokens-design-tokens--page#color), or a `url()` for an image
+   */
+  background?: BrandColor;
```
juanfabrega, line 16: To ensure consistency across these props in the Box and the CardFooter, I would change these to pull directly from BoxProps. E.G:

`background?: BoxProps['background'];`

src/components/Card/components/CardFooter/CardFooter.tsx
```diff
@@ -1,34 +1,60 @@
...
+  /**
+   * If defined as a prop, this value will take higher precedence than the corresponding component design token value
+   * Any valid [brand color token](/?path=/story/design-tokens-design-tokens--page#color) for the border color
+   * Or a responsive prop with BrandColor for each breakpoint.
+   */
+  borderColor?: BrandColor;
```
juanfabrega, line 22: Same comment as background.

juanfabrega (approved):
I had some minor comments but no showstoppers. 

I think a topic for later discussion is how this fits into the library moving forward, since it seems to complicate the class system for components. I can't think of other clear solutions at moment but doing this for larger components would create quite a bit of friction.

src/components/Card/Card.module.scss
nathanyoung: it is still necessary, but written this way

```
.card-section-border {
  + .card-section-border {
    border-top: 1px solid var(--card-section-border-color, var(--color-brand-grey-100));
  }
}
```

arempe93 (approved):

PR #752, opened 2022-04-11
feat: new Drawer UI component

nathanyoung:
# Github Issue or Trello Card
This PR addresses this issue: https://trello.com/c/Qza2RywS/1230-drawer-component

# Drawer component and documentation.

A Drawer is a panel that slides in from one edge of the viewport and overlays content on top of the page. It contains information or actions without leaving the context of the original page.

This leverages the same underlying [component](https://reach.tech/dialog/) as the `Modal` and `MediaModal` component to handle things like accessibility, aria principles, and focus management, but is styled to appear like a "drawer".

Drawer Example
Screen Recording 2022-04-14 at 10 22 57 AM

Placement Options
Screen Recording 2022-04-14 at 8 22 33 AM

- [x] Updating documentation.
- [ ] Updating Project Dependencies
- [ ] Improving or adding to Test Coverage

# Completeness Checklist

- [x] TESTS: My changes maintain the baseline required test coverage, as specified by code climate analysis.
- [x] DOCS: All new component work is covered in that component's `Overview` docs.
- [x] DOCS: All new component work is covered in a component `Playground`.
- [x] DOCS: All new visual changes or options are covered under relevant components' `VisualTests`.
- [x] My code has no linting or typescript compile warnings.
- [ ] My work is tied to a Github issue and satisfies the acceptance criteria (if applicable) of the corresponding issue.

# Quality Checklist:

- [x] I have added tests that prove my fix is effective or that my feature works

# UI Checklist


nathanyoung:
[Drawer component](https://trello.com/c/Qza2RywS/1230-drawer-component)

src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,236 @@
...
+      'position-absolute': containerRef !== undefined,
+    });
+
+    const contentClassnames = classNames(styles['drawer-content'], {
+      [styles.bottom]: placement === 'bottom',
+      [styles.left]: placement === 'left',
+      [styles.right]: placement === 'right',
+      [styles.top]: placement === 'top',
+      [styles['hide-overlay']]: hideOverlay,
+      'overflow-auto': !closeButton && !title,
```
nathanyoung, line 149: If the drawer has a header, don't apply scrolling to the entire content div so the header can be fixed.

yarn.lock
```diff
@@ -2264,10 +2264,10 @@
     react-remove-scroll "^2.4.3"
     tslib "^2.3.0"
 
-"@palmetto/palmetto-design-tokens@0.63.0":
```
nathanyoung, line 2267: changes in this file are due to the fact that we forgot to commit yarn.lock changes in other PRs where packages were updated.

src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,236 @@
...
+   */
+  title?: string;
+  /**
+   * The width of the Drawer when opened. Can be given a standard css value (px, rem, em, %),
+   * or a [width token](/?path=/story/design-tokens-design-tokens--page#width)
+   */
+  width?: DimensionSize | string;
+}
+
+export const Drawer: FC<DrawerProps> = forwardRef<HTMLDivElement, DrawerProps>(
```
dg-harris, line 101: Minor typing thing here, this actually isn't an `FC` because a functional component won't take refs.

I believe you can remove the explicit typing of `FC<DrawerProps>` and just let the type come from forwardRef

src/components/Drawer/Drawer.Overview.stories.mdx
```diff
@@ -0,0 +1,552 @@
...
+# Drawer
+
+A Drawer is a panel that slides in from one edge of the viewport and overlays content on top of the page. It contains information or actions without leaving the context of the original page.
+
+<Canvas withSource="open">
+  <Story
+    name="Basic Drawer"
+    parameters={{
+      chromatic: { disableSnapshot: true },
+    }}
```
nathanyoung, line 21: Tell chromatic to ignore these stories because the user has to click on the button to see the Drawer. Drawer.VisualTests.stories.tsx provides coverage of  Drawer snapshots

src/components/Drawer/Drawer.tsx
nathanyoung: @dg-harris thanks, didn't know that!

osbornm:
FWIW: the hidden overlay version does not move focus into the drawer, so you have to tab through the page to get into the drawer.

src/components/Drawer/Drawer.Overview.stories.mdx
RyanMacBern: If you want, you could programmatically click to open the drawer.
https://storybook.js.org/docs/react/writing-stories/play-function

nathanyoung: I tried this but it doesn't play nicely with the "Docs" tab and the code sample that is generated, so opted to rely on Drawer.VisualTests.stories.tsx to capture chromatic snapshots. Works great for the "Canvas" tab though.

Here's what I mean:
Screen Shot 2022-04-18 at 10 13 36 AM

src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,236 @@
...
+  className?: string;
+  /**
+   * Whether the drawer has a visible close button.
+   * If a title is defined, then a close button will be rendered
+   */
+  closeButton?: boolean;
+  /**
+   * If true, the drawer will close when the overlay is clicked
+   */
+  closeOnOverlayClick?: boolean;
```
RyanMacBern, line 47: By "overlay" are you referring to the drawer itself or the area outside the drawer? Should this be something like `closeOnClickOutside`?

src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,236 @@
...
+  /**
+   * Title to be displayed at the top of the Drawer.
+   * A close button will be rendered automatically if this prop is defined.
+   */
+  title?: string;
+  /**
+   * The width of the Drawer when opened. Can be given a standard css value (px, rem, em, %),
+   * or a [width token](/?path=/story/design-tokens-design-tokens--page#width)
+   */
+  width?: DimensionSize | string;
```
RyanMacBern, line 98: Should we lock this down beyond `string`?
```suggestion
   * The width of the Drawer when opened. Can be given a standard css value (px, rem, em, %),
   * or a [width token](/?path=/story/design-tokens-design-tokens--page#width)
   */
  width?: DimensionSize | `${number}${'px' | 'rem' | 'em' | '%'}`;
```

src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,236 @@
...
+  (
+    {
+      ariaLabel,
+      ariaLabelledBy,
+      allowPinchZoom = false,
+      children,
+      className,
+      closeButton = false,
+      closeOnOverlayClick = true,
+      containerRef = undefined,
```
RyanMacBern, line 111: I think you can remove all the `false` and `undefined` default values.

src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,236 @@
...
+
+    const renderHeader = () => {
+      if (closeButton && onDismiss && !title) {
+        return (
+          <Box
+            alignItems="flex-end"
+            justifyContent="center"
+            padding="md lg"
+            borderWidth="0 0 xs 0"
+            className={styles['drawer-header']}
```
RyanMacBern, line 161: Are we okay with splitting the styling across two places (Box props and CSS class)? My preference is to go with one or the other, but not both.

src/types/index.ts
```diff
@@ -63,6 +63,10 @@ export type StylesAndClasses<T> = {
 
 export type CssDimensionAbbreviation = 'h' | 'w' | 'mw' | 'mh' | 'minw' | 'minh';
 
+export type CssDimensionUnit = 'px' | 'rem' | 'em' | '%';
+
+export type CssDimensionValue = `${number}${CssDimensionUnit}`;
```
nathanyoung, line 68: this template string literal typescript type is causing the build to fail. tried to upgrade related ts packages but that didn't seem to resolve the issue, so I may just revert back to accepting strings like we do elsewhere.

src/components/Drawer/Drawer.tsx
nathanyoung: Box props is preferred since they are typed, but the `drawer-header` class enables theming via component design tokens (aka css vars). It is the only thing that is "theme-able" for this particular part of the component. We've done this in other content container components like Modal, MediaModal, and Card

src/components/Drawer/Drawer.tsx
nathanyoung: @juanfabrega i vaguely remember that we decided to put default values in most places so that it showed up in the props table documentation? Is that right?

nathanyoung: Oh hey, we actually documented that in the contribution guide!
See #3 here: https://github.com/palmetto/palmetto-components/blob/main/docs/CONTRIBUTING.md#component

src/types/index.ts
RyanMacBern: Looks like tsdx is dragging us down to Typescript v3.9.10. I'm playing around with the config to see if I can sort it out. I think this is also contributing to why I'm unable to symlink p-c into helios-app.

RyanMacBern: I'm making some progress by swapping `tsdx` (abandoned before Typescript v4) with `dts-cli` (forked from tsdx to add Typescript v4 support). This essentially upgrades the whole project from Typescript v3 to Typescript v4, so I have several things to sort out. I'll post a PR when it's ready.

src/components/Drawer/Drawer.tsx
RyanMacBern: It just says "Defaults for optional props are declared in the function declaration for the component." Nothing about showing up in documentation.

src/components/Drawer/Drawer.tsx
RyanMacBern: If we update the type for `borderColor` to `borderColor?: BrandColor | `var(${string})`;`, then we could do
```suggestion
          <Box
            alignItems="flex-end"
            justifyContent="center"
            padding="md lg"
            borderWidth="0 0 xs 0"
            borderColor="var(--drawer-header-border-color, var(--color-brand-grey-100))"
```

nathanyoung: That would impact the `<Box>` and components using this pattern, so prefer to keep suggested change out of this scope and approach it holistically.

RyanMacBern (approved):
👍

src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,234 @@
...
+      : width;
+
+    const dynamicStyle: CSSProperties = {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      ['--w' as any]: dynamicWidth,
+    };
+
+    const overlayClassnames = classNames(styles.overlay, styles.drawer, {
+      [styles['hide-overlay']]: hideOverlay,
+      [styles['hide-overlay-left']]: hideOverlay && placement === 'left',
```
juanfabrega, line 133: Minor style thing but iff you wanted you could write these 4 lines as 1 -->
```
[styles[`hide-overlay-${placement}`]: hideOverlay,
```

src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,234 @@
...
+      [styles['hide-overlay-right']]: hideOverlay && placement === 'right',
+      [styles['hide-overlay-bottom']]: hideOverlay && placement === 'bottom',
+      [styles['hide-overlay-top']]: hideOverlay && placement === 'top',
+      'position-fixed': containerRef === undefined,
+      'position-absolute': containerRef !== undefined,
+    });
+
+    const contentClassnames = classNames(styles['drawer-content'], {
+      [styles.bottom]: placement === 'bottom',
```
juanfabrega, line 142: If you wanted you could condense these 4 lines into 
```
styles[placement],
```

src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,234 @@
...
+import classNames from 'classnames';
+import { DimensionSize, CssDimensionValue } from '../../types';
+import { WIDTH_OPTIONS } from '../../lib/tokens';
+import { Box } from '../Box/Box';
+import { Icon } from '../Icon/Icon';
+import styles from './Drawer.module.scss';
+
+export type DrawerPlacementType = 'left' | 'right' | 'top' | 'bottom';
+
+export const DRAWER_PLACEMENT: DrawerPlacementType[] = ['right', 'left', 'top', 'bottom'];
```
juanfabrega, line 12: This export is only used internally for our testing/storybook it seems. Might be best to just move it out of the component altogether unless we want to export it to consumers.

src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,234 @@
...
+      containerRef = undefined,
+      dangerouslyBypassFocusLock = false,
+      dangerouslyBypassScrollLock = false,
+      hideOverlay = false,
+      initialFocusRef = undefined,
+      isOpen,
+      onDismiss = undefined,
+      placement = 'right',
+      title = undefined,
+      width = undefined,
```
juanfabrega, line 118: we probably want to spread props here for consumer flexibility and things such as data attributes, etc. I am not sure if reach UI gives us a prop type to extend though. Maybe something we can add after an initial merge since it is not make or break.

juanfabrega (approved):

PR #823, opened 2023-04-20
Allow OptionTileGroup options to have individual errors

arempe93:
Now, only the "Loanpal 20 Year 3.99%" appears as if it has an error

image

image


arempe93:
was going to try to fix the focus ring colors as well, but couldn't find the right variable

image

could you help me on that tomorrow @nathanyoung ?

nathanyoung:
Absolutely. We may have to create a new variable for this state. It’s been
a while since I’ve looked at this.

nathanyoung (approved):

